Issue #6, opened 2015-03-25
Currying is useful, but pretty unexpected


nhunzaker:
It is very easy to miss that currying exists. I propose a `curry` or `prepare` method.

PR #7, opened 2015-03-26
[change] Replace currying with prep. Add uniqueness check to stores

### LINKED ISSUE #5: Throw an error if a Store's toString method is not unique

nhunzaker:
Fixes #5 and #6.

The big change here is the removal of automatic currying of actions sent into an app. This was super neat, but too novel. I was quickly confused when a callback wasn't firing (I was expecting the default argument of the action to be respected).

In light of this, I've replaced this functionality with an `app.prepare` function. This partially applies an action, which is really what I wanted in the first place.

---

Additionally, Microcosm will now throw an error if a Store does not stringify to a unique name.


zporter:
:heart: :+1:

src/assign.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * assign
+ * Non-destructively assign one object into another
+ */
+
+export default function assign (a, b) {
+  return { ...a, ...b }
```
greypants: r u from teh future?

greypants: so will a's properties be overwritten with b's or no?

greypants:
:+1:

src/assign.js
nhunzaker: This spreads the properties of `a` and `b` into a new object. So it safely merges the two objects without mutating either of them.

nhunzaker: Within Babel, this is basically:

before:

```javascript
let c = { ...a, ...b }
```

after:

```javascript
var _extends = require('babel/path/to/extend')
var c = _extends({}, a, b);
```

greypants: cool.

nhunzaker: Yeah! I love it. I would have done it more inline within the code but I thought this was a bit clearer

averyvery: :+1:

PR #12, opened 2015-03-30
Plugins

nhunzaker:
I needed to run environment specific behavior before an app is run. I did this by adding a simple, queued, plugin system and a start() method that will wait for plugins to finish before booting an app.

The other big breaking change here is that application state is not calculated until `start` is called. In addition to reducing some complexity, this also means `undo` history can anchor itself upon this initial start value.


solomonhawk:
:+1:

nhunzaker:
Also worth noting, this feature has been specifically designed to help solve two real problems:

1. For Pearl, I want to prefetch data for an application on the server only, background loading new content in the client
2. On Colonel Kurtz, I want the ability to add a "data migration" plugin that will automatically update old schemas.

src/install.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Install
+ * A simple FIFO queue for installing plugins
```
greypants: what the what is a FIFO queue?

nhunzaker: First in, first out!

src/install.js
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+export default function install ([plugin, ...tail], app, callback) {
+  if (!plugin) return callback()
+
+  plugin.register(app, app._options, function(err) {
+    if (err) throw err
+    install(tail, app, callback)
```
greypants: I'm not following what's happening here... when you run install, plugin.register gets called... which also calls install in its callback?

greypants:
¯\\(°_o):thumbsup:

src/install.js
nhunzaker: Basically, run a plugin, when it finishes check if any others exist, then run those. This works sort of like `.then` with promises, but Promises are terrible about hiding errors and I'd like to avoid them where possible.

zporter: Recursion (calls `install` with the `tail`)

src/__tests__/plugin-test.js
```diff
@@ -0,0 +1,43 @@
...
+describe('Microcosm', function() {
+
+  it ('can add plugins', function(done) {
+    let m = new Microcosm()
+
+    m.addPlugin({
+      register: function(app, options, next) {
+        next()
+      }
+    })
```
averyvery: wish you could write this stuff and use babel (easily, at least)

nhunzaker: I should be using babel here :( Do you mean shorthand notation?

src/install.js
averyvery: I didn't get this at first but now I see — you're recursively calling install for each plugin in the plugin array. Is there a need for the recursion, or is it just fun?

src/__tests__/plugin-test.js
averyvery: I meant, you're not using `() =>` or `register(app, options, next)` signatures. I assumed it was because running babel on tests is a pain, but looks like I'm wrong.

src/install.js
nhunzaker: @averyvery I want them to install happen sequentially, regardless of async

nhunzaker: Sorry. The alternative was to use Promises, but it was heavy handed and Promises are terrible about hiding errors

src/__tests__/plugin-test.js
nhunzaker: No, I just forgot. Haha.

nhunzaker: Worth noting. These tests receive the same processing as the main source code. They all pass through the same webpack configuration.

src/install.js
greypants: ooooh. i see how it works now. Interesting.

PR #17, opened 2015-04-01
Cleanliness

nhunzaker:
Basic refactor of tests, but I also added a secondary argument to `pull` to help clean up some queries in Colonel.


src/Microcosm.js
```diff
@@ -25,8 +27,9 @@ export default class Microcosm {
     this.commit(this.deserialize(data))
   }
 
-  pull(key) {
-    return this._state[key]
+  pull(key, fn) {
+    let val = this._state[key]
+    return typeof fn === 'function' ? fn.call(this, val) : val
```
nhunzaker: Hmm. Scope should not matter here. What about

```
pull(key, fn, ...args) {
//...
fn(..args)
}
```

nhunzaker: The specific use case is:

```
let blockType = app.pull('blockTypes', findBy, block.type)
```

PR #18, opened 2015-04-02
Dev 6.1.0

nhunzaker:
`Microcosm.pull` lets you quickly retrieve nested keys


src/getIn.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * getIn
+ * Enumerate through a list of keys to get a value
+ */
+
+export default function getIn (obj, key) {
+  let val  = obj
```
greypants: y the name reassignment?

greypants:
:+1:

src/getIn.js
nhunzaker: I don't even know. Good idea

nhunzaker: Done

PR #23, opened 2015-05-11
Multiple action formats

nhunzaker:
Actions once again support promises and returned values. We strictly enforce parameters passed to arguments to be one function, however this should help with the upgrade. I'm comfortable making this a minor release given it does not break actions period, it only makes their usage more strict.

This means you can design actions to follow:

**Return a value:**

```javascript
action: function(params) {
    return params
}
```

**Return a promise:**

```javascript
action: function(params) {
    return new Promise(...)
}
```

**Provide an error-first callback:**

```javascript
action: function(params, next) {
  if (!params.required) {
    next(new Error('Please provide the required field'), null)
  } else {
    next(null, 'stuff')
  }
}
```


src/Microcosm.js
```diff
@@ -73,20 +74,17 @@ class Microcosm extends Foliage {
     return this
   }
 
-  prepare(action, ...buffer) {
+  prepare(/* action, ...buffer */) {
```
solomonhawk: Forgive the ignorance, what change/benefit is there to doing it this way in comparison to rest params as you had before?

src/Signal.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    callbacks.forEach(fn => fn(error, result))
+  })
+}
+
+function passThrough (action, params, next, callbacks) {
+  let result = action(params)
+
+  // Actions some times return thenables. When this happens, wait for
```
solomonhawk: `thenable` :+1:

solomonhawk:
:ship:

src/Microcosm.js
nhunzaker: It's silly, but the compiled code is smaller. I'll run the numbers, the confusion is not worth 20 bytes

nhunzaker: Uhhh. oddly. The difference is 17 bytes. I'll just keep the spread.

src/Signal.js
nhunzaker: Yeah! We support the Promise API without expecting promises to exist

PR #36, opened 2015-06-01
Fix store scope

nhunzaker:
cc @averyvery 

This PR makes all dispatched store actions be called within the scope of the store itself.


averyvery:
:+1:

PR #41, opened 2015-06-12
Store refactor

nhunzaker:
This is mostly internal changes. However it highlights the beginning of trying to reduce the surface area of stores.

The biggest change here is how Stores accept actions:

```javascript
// Before
store.send(state, action, payload)
//After
Store.send(store, action, state, payload)
```

So now, this is a static method of `Store`. `send` is no longer on the prototype.

---

Summary of work:

1. Upgrade Foliage to 0.21.0
2. Move Store.prototype.send to Store.send. Change function signature to align more with other methods
3. Add `babel-plugin-object-assign` so that Object.assign can be used under the hood (transpiled polyfill)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,5 +3,6 @@
   "loose": true,
   "sourceMaps": "inline",
   "blacklist" : [ "strict" ],
-  "optional"  : []
+  "optional"  : [],
+  "plugins": [ "object-assign" ]
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## 8.2.0
+
+## Internal Changes
+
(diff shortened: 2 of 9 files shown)


.babelrc
```diff
@@ -3,9 +3,6 @@
   "loose": true,
   "sourceMaps": "inline",
   "blacklist" : [ "strict" ],
-  "optional"  : [
-    "minification.inlineExpressions",
-    "minification.propertyLiterals",
-    "minification.memberExpressionLiterals"
-  ]
+  "optional"  : [],
```
nhunzaker: These optimizations yielded little benefit (if any) and threw a bunch of babel warnings

averyvery:
:+1: interesting idea. Fewer imports required on components with multi-store effects, I suppose.

PR #45, opened 2015-06-22
Transactional

nhunzaker:
This PR rewrites the Microcosm internal data layer to use a transactional system. This should support optimistic updates and future efforts on undo/redo functionality. There is no longer a need for Foliage in core, so I removed it. 

### In short:

1) Adds transactions. Microcosm has a `base` and `state` property. `base` represents the reduction of all completed transactions. `state` represents `base` state + pending transactions (such as optimistic updates with a pending Promise representing persistence with an API).
2) Removed Foliage. As a fun aside: Microcosm is about 0.75kb smaller.
3) Removes Store "wrapper" class in favor of pure functions
4) `start()` no longer throws an error and instead forwards error to provided callbacks. This should make it easier to respond to errors in a more intelligent way.
5) Removes `get` and `set`, instead just reference `this.state`

### Outstanding items:

- [ ] Documentation for `9.0.0`
- [x] Audit tests. Are we covering all edge cases? Are any tests silly coverage chores?
- [ ] Does the nomenclature feel right?


nhunzaker:
Just for posterity. We did an internal code review here in substitute of :+1: s

PR #46, opened 2015-06-24
Clean up transactions by making class

nhunzaker:
Signal felt pretty gnarly, so I cleaned it up a bit my moving some of the logic out into a `Transaction` class. I also added "states" to Transactions to make it clearer what is going on.

### DIFF
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -7,12 +7,12 @@
  * in one place.
  */
 
-let Diode   = require('diode')
-let Store   = require('./Store')
-let install = require('./install')
-let plugin  = require('./plugin')
-let remap   = require('./remap')
-let signal  = require('./signal')
+let Diode       = require('diode')
+let Store       = require('./Store')
+let Transaction = require('./Transaction')
+let install     = require('./install')
+let plugin      = require('./plugin')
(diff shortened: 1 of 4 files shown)


src/Microcosm.js
```diff
@@ -114,26 +111,13 @@ Microcosm.prototype = {
...
 
-      transaction.done = transaction.done || done
-
-      this.rollforward()
+    this.transactions.push(transaction)
 
-      return body
-    }, action.apply(this, params))
+    return transaction.run(params)
   },
```
mzlock: such spaces between everything

nhunzaker: Such read, :book:

mzlock:
:+1:

src/Microcosm.js
nhunzaker: Actually this still feels a bit awkward. Transactions are event emitters, but they only have one event. I had considered just doing:

`new Transaction(action, params, this.rollforward.bind(this)`

nhunzaker: Maybe I should just do that...

nhunzaker: Well hmm. That's equally as awkward. I guess I'll leave it alone for now.

PR #48, opened 2015-06-30
v9.x callbacks

nhunzaker:
`app.push` now conducts the following behavior:

```javascript
// app.push(action, arguments, callback)
app.push(action, [ 'one', 'two', 'three' ], function(error, body) {
  // error of failed iteration
  // body of last iteration
})
```

I also picked up a few tricks from https://github.com/then/promise to ensure errors that occur in the callback do not get trapped by promise chains.

Also... I refactored `signal.js`, it feels quite a bit cleaner

### DIFF
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -134,8 +134,8 @@ Microcosm.prototype = {
    * Clear all outstanding transactions and assign base state
    * to a given object (or getInitialState())
    */
-  reset(state=this.getInitialState(), transactions=[]) {
-    this.transactions = transactions
+  reset(state=this.getInitialState()) {
+    this.transactions = []
     this.base = state
 
     return this.transact()
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -107,8 +107,8 @@ Microcosm.prototype = {
   /**
    * Partially applies `push`.
    */
-  prepare(action, ...params) {
-    return this.push.bind(this, action, ...params)
+  prepare(action, params) {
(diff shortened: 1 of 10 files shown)


src/Microcosm.js
```diff
@@ -107,8 +107,8 @@ Microcosm.prototype = {
   /**
    * Partially applies `push`.
    */
-  prepare(action, ...params) {
-    return this.push.bind(this, action, ...params)
+  prepare(action, params) {
+    return (more, callback) => this.push(action, [].concat(params, more), callback)
   },
```
nhunzaker: `prepare` had to change because we no longer have an infinite argument length for `push`. The new signature is:

```javascript
let summer = app.prepare(sum, [ 1, 2 ])

summer(1, function(error, body) {
  assert.equal(body, 4)
})

summer([ 1, 1 ], function(error, body) {
  assert.equal(body, 5)
})

File: src/Microcosm.js [hunk 2]
```diff
@@ -134,8 +134,8 @@ Microcosm.prototype = {
    * Clear all outstanding transactions and assign base state
    * to a given object (or getInitialState())
    */
-  reset(state=this.getInitialState(), transactions=[]) {
-    this.transactions = transactions
+  reset(state=this.getInitialState()) {
+    this.transactions = []
```
nhunzaker: Not related to this PR, but I went a head and removed the default transaction list argument from `reset`. Eventually, I'd like to be able to inject history when you reset so that you can rehydrate an app with "undo" state.

greypants:
:+1:

PR #50, opened 2015-07-07
Replace optimistic updates with chat example

nhunzaker:
I added a chat example using the [Eliza](http://nlp-addiction.com/eliza/) chat bot. The goal of this example is to demonstrate optimistic updates.

chat


examples/chatbot/actions/sendChat.js
```diff
@@ -0,0 +1,34 @@
...
+ * `sendChat` is an ES6 generator. Generators use the `yield` keyword to create
+ * steps in an ES6 iterable.
+ *
+ * Microcosm processes each of these iterations sequentially, using the most
+ * recently resolved iteration as the payload of the action.
+ *
+ * More on generators:
+ * http://babeljs.io/docs/learn-es2015/#generators
+ */
+export default function* sendChat (message) {
```
ltk: what does the `*` do?

nhunzaker: It indicates that the function is a generator

solomonhawk: It's a generator. https://babeljs.io/docs/learn-es2015/#generators

nhunzaker: Here's a more concrete example:

```javascript
function* range(start, end) {
  
  while (start <= end) {
    yield start++
  }
  
}


for (let n of range(0, 10)) {
  console.log(n) // 0,1,2,3,4,5,6,7,8,9,10
}
```

Give that a go here: http://babeljs.io/repl/

solomonhawk:
:+1:

examples/chatbot/components/conversation.jsx
```diff
@@ -0,0 +1,32 @@
...
+  getDefaultProps() {
+    return {
+      messages: []
+    }
+  },
+
+  componentDidUpdate() {
+    let el = React.findDOMNode(this)
+
+    el.scrollTop = el.scrollHeight
```
averyvery: Ah, this is nice.

examples/chatbot/components/conversation.jsx
```diff
@@ -0,0 +1,32 @@
...
+  render() {
+    const { messages } = this.props
+
+    return (
+      <ol className="conversation">
+        { messages.map(this.getMessage) }
+      </ol>
+    )
+  }
+
```
averyvery: This might be overkill, but it would be fun to add a little a11y with an aria announcement for chat. Meh example: https://github.com/vigetlabs/know-lupus-game/blob/master/front-end/src/assets/js/components/Banner/Announcements.js

averyvery:
:+1:

examples/chatbot/components/conversation.jsx
nhunzaker: Oh, that's a fantastic idea. These examples are educational, no reason we can't teach other things besides Microcosm.

nhunzaker: Would you just announce the latest reply?

averyvery: Yeah, I think that's the only relevant info. You can look at how Slack does it by running `document.querySelectorAll('#aria_live_announcer')` on a slack.com page — they just announce the room change and any new messages.

nhunzaker: For non-vigeters, this is what Doug's talking about:

```javascript
module.exports = React.createClass({
  render() {
    return <div className="screenreader-only" aria-live="assertive">
      {this.props.announcement}
    </div>
  }
})

nhunzaker: For posterity: #51

PR #52, opened 2015-07-10
Reduce some array allocations

nhunzaker:
We do some work in places that we don't have to. This PR reduces the array allocation count of the transaction lifecycle tremendously. It also makes `clean` more granular, which will help with lifecycle methods later down the road.

### DIFF
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -1,10 +1,10 @@
 let Diode = require('diode')
 let Store = require('./Store')
 let Transaction = require('./Transaction')
+let flatten = require('./flatten')
 let install = require('./install')
 let plugin = require('./plugin')
 let remap = require('./remap')
-let flatten = require('./flatten')
 
 let noop = a => a
 
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -4,6 +4,9 @@ let Transaction = require('./Transaction')
 let install = require('./install')
 let plugin = require('./plugin')
 let remap = require('./remap')
+let flatten = require('./flatten')
+
+let noop = a => a
 
 let Microcosm = function() {
   /**
@@ -56,8 +59,8 @@ Microcosm.prototype = {
   /**
    * Used to determine if a transaction should be purged during `clean()`
    */
-  transactionIsValid(item) {
-    return !item.error
+  shouldRejectTransaction(item) {
+    return item.error
   },
 
   /*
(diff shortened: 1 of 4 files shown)


src/flatten.js
```diff
@@ -0,0 +1 @@
+module.exports = Array.prototype.concat.bind([])
```
nhunzaker: This let's us use `[].concat` to ensure arguments passed into `transaction`, `push` and `prepare` are arrays. There's no reason to use a new array every single time.

src/Transaction.js
```diff
@@ -23,27 +18,27 @@ Transaction.prototype = {
...
-    // https://github.com/then/promise/blob/master/src/done.js
-    setTimeout(this.callback.bind(this, error, body), 0)
-  },
-
   isComplete() {
     return this.error || this.done
   },
 
-  run() {
+  run(resolve, reject, progress) {
```
nhunzaker: I wanted this to follow the signature of Promises (using the old progress callback from the A specification)

src/Microcosm.js
```diff
@@ -101,12 +105,10 @@ Microcosm.prototype = {
...
    * Engages the transaction life cycle.
    *
-   * 1. Clean up erroneous transactions
-   * 2. Squash down complete transactions
-   * 3. Roll outstanding changes forward into new state
+   * 1. Squash down complete transactions
+   * 2. Roll outstanding changes forward into new state
    */
   transact() {
-    this.clean()
```
nhunzaker: We don't need to clean every time, only when a transaction is rejected.

tommymarshall:
:+1:

tommymarshall:
:+1:

src/Microcosm.js
```diff
@@ -70,8 +73,10 @@ Microcosm.prototype = {
   /**
    * Remove invalid transactions
    */
-  clean() {
-    this.transactions = this.transactions.filter(this.transactionIsValid)
+  clean(transaction) {
+    if (this.shouldRejectTransaction(transaction)) {
+      this.transactions = this.transactions.filter(t => t !== transaction)
```
averyvery: classy

averyvery:
:+1:

src/flatten.js
tommymarshall: neat

tommymarshall:
:+1:

PR #53, opened 2015-07-13
Refactor plugins for lifecycle methods

nhunzaker:
This is a change from an outstanding branch that adds lifecycle methods. Not much, but plugins were instantiated on `start` before, which is problematic for the `appWillStart` lifecycle method :).


tommymarshall:
:+1:

PR #54, opened 2015-07-14
Transaction refactor

nhunzaker:
I'm starting to write the Microcosm dev tools and I noticed a couple of issues with our approach. Specifically, when I send transactions over `window.postMessage` to the Chrome extension, it includes a reference to the action type as a function. This is a problem when the browser attempts to serialize the object using this algorithm:

https://developer.mozilla.org/en-US/docs/Web/Guide/API/DOM/The_structured_clone_algorithm

I still do not have an answer yet for `Error` objects returned by promises. They also don't serialize (which is really unfortunate). Still, this lays the groundwork.

### DIFF
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -7,8 +7,6 @@ let plugin = require('./plugin')
 let remap = require('./remap')
 let tag = require('./tag')
 
-const EMPTY_ARRAY = []
-
 let Microcosm = function() {
   /**
    * Microcosm uses Diode for event emission. Diode is an event emitter
@@ -150,7 +148,7 @@ Microcosm.prototype = {
    * Clear all outstanding transactions and assign base state
    * to a given object (or getInitialState())
    */
-  reset(state=this.getInitialState(), transactions=EMPTY_ARRAY) {
+  reset(state=this.getInitialState(), transactions=[]) {
     this.transactions = flatten(transactions)
     this.base = state
 
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -7,6 +7,8 @@ let plugin = require('./plugin')
 let remap = require('./remap')
 let tag = require('./tag')
 
+const EMPTY_ARRAY = []
+
 let Microcosm = function() {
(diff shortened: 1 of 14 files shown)


examples/simple-svg/components/Viget.jsx
```diff
@@ -4,7 +4,7 @@ export default React.createClass({
   getDefaultProps() {
     return {
       title: 'Microcosm SVG Chart Example',
-      height: 250,
+      height: 300,
       width: 250
     }
   },
```
nhunzaker: Probably shouldn't be in here, but the clipping was bugging me :)

src/Store.js
```diff
@@ -3,35 +3,22 @@
...
 
-    return attempt(Store.register(store), action.toString(), [ state, body ], state, store)
-  }
+exports.send = function (store, state, { payload, type }) {
+  let handler = 'register' in store ? store.register()[type] : false
 
+  return handler ? handler.call(store, state, payload) : state
 }
-
-module.exports = Store
```
nhunzaker: I dropped `attempt`. It was super hard to parse and, while this is verbose, I like the clarity

src/Microcosm.js
```diff
@@ -75,7 +74,7 @@ Microcosm.prototype = {
    */
   clean(transaction) {
     if (this.shouldRejectTransaction(transaction)) {
-      this.transactions = this.transactions.filter(t => t !== transaction)
+      this.transactions.splice(this.transactions.indexOf(transaction), 1)
```
nhunzaker: Small performance thing. This helps quite a bit when displaying high frequency data visualizations on mobile

src/plugin.js
```diff
@@ -13,7 +13,7 @@ const defaults = {
   },
 
   register(app, options, next) {
-    next()
+    next(null)
```
nhunzaker: Just in case someone checks `arguments.length` for errors. Who knows?

ltk:
:+1::+1:

PR #57, opened 2015-08-06
Add ability to set payload on Transaction Create. FSA test

nhunzaker:
Not too much to report here. I wanted the ability to set a payload when creating a transaction and I added the [flux-standard-action](https://github.com/acdlite/flux-standard-action) standard test into the suite.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,12 @@
   "babel": {
     "stage": 1,
     "sourceMaps": "inline",
-    "blacklist": [ "strict" ],
-    "plugins": [ "object-assign" ]
+    "blacklist": [
+      "strict"
+    ],
+    "plugins": [
+      "object-assign"
+    ]
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


cwmanning:
What's a payload in this context?

nhunzaker:
@cwmanning I was a bit cryptic, wasn't I? It is the data associated with a dispatched action. It is what is forwarded to Stores listening to specific actions.

nhunzaker:
So like if you were to dispatch an action that increases a step count by a given value, that might look like:

```
{
    type: 'step',
    payload: 2
}
```

And your store would respond with something like

```javascript
function increaseStep(currentCount, amount) {
    // Amount here is the payload of the action
    return currentCount + amount
}
```

cwmanning:
Oh gotcha. Didn't realize from the diff alone that payload is already a thing here.

:+1:

PR #58, opened 2015-08-12
Add life cycle methods for transactions

nhunzaker:
This commit renames and tweaks the update and error callbacks used by
Microcosm to process transactions to conform to more of a lifecycle
method of operation. This is a step towards actual lifecycle methods.


cwmanning:
:+1:

PR #59, opened 2015-08-21
Add experimental microcontroller mixin

nhunzaker:
This is the beginning of a ["getters"](https://github.com/optimizely/nuclear-js#performance) or ["selectors"](https://github.com/faassen/reselect) like feature for Microcosm. It will probably never be as robust as NuclearJS's getters, which can leverage ImmutableJS hash codes, but it can at least aspire to improve efficiency.

This is just the subscription system. For actual selection, we could actually use https://github.com/faassen/reselect


addons/microcontroller/index.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * Microcontroller
```
ltk: I think you could go for a more descriptive name for this. controller is already a kind of loaded word in software.

nhunzaker: Agreed. I'd actually like something much more concise. Let's brainstorm!

Makefile
```diff
@@ -33,6 +33,9 @@ javascript-min: javascript
 release: clean build
 	npm publish $(DIST)
 
+release-beta: clean build
```
greypants: ooh. these are handy.

addons/microcontroller/index.js
greypants: Why not `ComputedProperties`? Also, I feel like `selections` should also be `computedProperties`. It better describes them, and would link it to the mixin, so it's clear they go with each other.

greypants:
:+1:

addons/microcontroller/index.js
averyvery: `helpComputer`

`stop all the downloading`

addons/microcontroller/index.js
```diff
@@ -0,0 +1,77 @@
...
+  },
+
+  componentDidMount() {
+    const instance = this._getInstance()
+
+    if (!instance) {
+      throw TypeError(`Microcontroller expected ${this.constructor.displayName} to have an "app" or "instance" property but got ${ typeof instance }`)
+    }
+
+    this._getInstance().listen(this._change)
```
averyvery: Since you already have instance, why use `_getInstance` again?

averyvery:
:+1:

addons/microcontroller/index.js
nhunzaker: Reasons, Doug. Reasons... 

I'll fix it

nhunzaker:
Great feedback. Closing for now until the name thing can be sorted out.

PR #62, opened 2015-08-31
Refactor plugin install process

nhunzaker:
This commit uses a more straightforward queue when installing plugins and removes the need to add a `__start` method to plugins.


src/plugin.js
```diff
@@ -7,16 +7,25 @@
...
+  return function (error) {
+    // Halt execution of all future plugin installation if there is an error
+    if (error) {
+      return next(error)
+    }
 
-  register(app, options, next) {
-    next(null)
+    // Plugins might not have a register method. In this case, just continue through
+    return plugin.register ? plugin.register(plugin.app, plugin.options, next) : next(null)
```
solomonhawk: Check for `typeof === 'function'` perhaps?

solomonhawk:
:+1:

src/plugin.js
nhunzaker: Hmmm. That would kind of be a silent error though. What about a dev-only error?

jeremyfrank:
:+1:

src/plugin.js
nhunzaker: Thoughts on ca2747cd5c144af35049f4432d193568b3929199?

solomonhawk: Looks good

PR #63, opened 2015-09-01
Keep stores as 2D arrays

nhunzaker:
We made guarantees about the processing order of stores. That
whenever an action is dispatched, stores will be processed in
the order with which they are added.

Unfortunately, before this commit we kept these stores in a POJO.
While many browsers treat objects as ordered maps _most_ of the time,
the cost to just use a 2D array is negligeable and allows us to drop
the `remap` module in favor of the more intuitive `Array.prototype.reduce`.

Additionally, it allows stores to be enumerated over using iterators like:

```javascript
for (let [ key, store ] of app.stores) {}
```

Or using an ES6 Map like:

```javascript
new Map(app.stores).forEach
```

And (the most important :P), it's actually 2 bytes smaller than the previous implementation. PURE JOY.


ltk:
real good. nice :+1: for you

PR #64, opened 2015-09-01
Stores can have a diversity of registration methods.

nhunzaker:
This commit allows the following behaviors:

```javascript
function Store() {
  return {
    [action]: (a,b) => a + b
  }
}
```

```javascript
function Store() {
  return {
    [action]: 10
  }
}
```

This is to support some future work on refactoring the store, dispatching process. It also opens up the flexibility for stores to interact with other parts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 9.3.0
+
+### Noticeable changes
+
+- Store registration methods can return non-function values. When this
+  is the case, it will use this value as the new state.
+
 ## 9.2.0
 
 ### Noticeable changes
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "microcosm",
-  "version": "9.2.0",
+  "version": "9.3.0",
   "private": true,
   "description": "A variant of Facebook's Flux with centralized, isolated state",
   "main": "src/Microcosm.js",
--- a/src/Store.js
+++ b/src/Store.js
@@ -27,7 +27,7 @@ exports.deserialize = function (store, raw) {
 }
 
 exports.send = function (store, state, { payload, type }) {
-  let pool = store.register? store.register() : null
+  let pool = typeof store.register === 'function' ? store.register() : null
 
   if (!pool || type in pool === false) {
     return state
--- a/src/Store.js
+++ b/src/Store.js
@@ -35,7 +35,7 @@ exports.send = function (store, state, { payload, type }) {
     pool = store.register()
   }
 
-  if (pool == undefined || type in pool === false) {
+  if (pool == null || type in pool === false) {
     return state
   }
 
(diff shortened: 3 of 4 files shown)


ltk:
why would you want a handler to not be a function? slash why would you want to use actions/the dispatcher for things actions that just return a constant?

Also it feels like letting stores be functions is kinda needless syntactic sugar that just increase the complexity of this thing. But you probably have a vision that I just can't see yet.

nhunzaker:
@ltk Ooo debate time.

The idea was to allow Stores to act like switches that can turn on and off values. For example:


```javascript
function TrafficLight() {
    return {
        [red]      : 'stop',
        [yellow] : 'caution',
        [green]  : 'go'
    }
}
```

I see two draw backs to handlers not being functions. One, if a bad reference to a method is given `(this.doesNotExist)`, it'll return undefined. Also, it adds another way to configure them, possibly increasing the learning curve (but not really enough that I'm concerned about it).

What are your feels here?

As for allowing Stores to be functions. The big idea is to allow them to be defined one level higher than a private object. Instead, the module itself is the store. So instead of exporting a Store object, you configure the register method as the public interface of the default export:

Consider a Store as an object:

```javascript
const Planets = {
    getInitialState() {
        return []
    },
    register() {
        return {
            [addPlanet]: Planets.add
        }
    },

    add(planets, props) {
        return planets.concat(props)
    }
}

export default Planets
```

I have a couple of questions here. What value is having an object providing? `add` is a pure function - what value is it adding being a member of Planets?

Why build an object in the first place, when we have a similar pattern already given to us via modules? I'm proposing a Store to look like:

```javascript
function getInitialState () {
    return []
}

function add (planets, props) {
    return planets.concat(props)
}

function register() {
    return {
        [addPlanet]: add
    }
}

export default { register, getInitialState }
```

So this cuts the object, arguable that it's _that_ much better. However there is a step beyond this in my roadmap. I want to drop `serialize`, `deserialize`, and `getInitialState` as "special" methods and just make them _lifecycle methods_. So the final form (just like Dragon Ball Z, there is always _another_ final form, but anyway):

```javascript
import Microcosm from 'microcosm'

const EMPTY = []

export function add (planets, props) {
    return planets.concat(props)
}

export default function register() {
    return {
        [Microcosm.willStart] : EMPTY,
        [addPlanet]                : add
    }
}
```

Those are my thoughts anyway. Thoughts?

ltk:
okay, interesting idea. so you could have more granular stores that control specific values instead of fewer stores that transform much larger state objects. i dig it.

I'm still not sold either way on stores as functions. It seems a little too opinionated to always assume that   stores will be defined within modules, and it doesn't seem worth the complexity to support both ways of doing this. My only strong feel is that it probably should be one way or the other.

nhunzaker:
@ltk Fair point. Stores as functions is a design decision for future work. I'll remove it out of the context of this PR so that we can evaluate it when we get to that point.

nhunzaker:
@ltk Oh I forgot! One final note. Plugins can now be functions, so there would be a similar pattern there. In any case, let's add lifecycle actions first before we move to that.

nhunzaker:
@ltk Removed stores as functions. We'll cross that bridge when we get there.

ltk:
:+1:

PR #65, opened 2015-09-02
A plugin can be a simple function

nhunzaker:
In the same vein as stores, plugins can now simply be functions:

so given:

```javascript
let Plugin = {
    register(app, options, next) {
        // do stuff
        next()
    }
}
```

you can additionally do:

```javascript
function register(app, options, next) {
    // do stuff
    next()
}
```
I also refactored the plugin storage mechanism a bit. Oddly enough, this actually reduces the build size as well.


src/plugin.js
```diff
@@ -7,17 +7,13 @@
...
  */
 
-function PluginFactory (config, options, app) {
-  return Object.assign({ app, options }, config)
-}
-
 function checkPlugin (plugin) {
-  if (process.env.NODE_ENV === 'development' && 'register' in plugin && typeof plugin.register !== 'function') {
-    throw TypeError('Expected register property of plugin to be a function, instead got ' + plugin.register)
+  if (process.env.NODE_ENV !== 'production' && typeof plugin !== 'function' && ('register' in plugin && typeof plugin.register !== 'function')) {
```
ltk: Unrelated to this PR, and I'm working on a thing now related to this, but I don't like these types of `env` checks in all the places. The app should have a `debug` mode that could optional be configured to equal `process.env.NODE_ENV !== 'production'`, but alternatively could be anything our hearts desire.

Just a random thought.

src/plugin.js
```diff
@@ -7,17 +7,13 @@
...
+    throw TypeError('Expected plugin to be a function or entity with a method property.')
```
ltk: this `method property`isn't super clear. You just need a function defined as the property `register`, right?

ltk:
looks gud :+1:

src/plugin.js
nhunzaker: Me too. I don't like it either. Hmmm

nhunzaker: I'll send out a PR after this with some ideas. I need to benchmark some builds.

src/plugin.js
nhunzaker: How does `Microcosm plugins can either be objects with a register method, or functions that follow the same pattern.` sound?

nhunzaker: (updated)

nhunzaker: This is dev build only, so we could also do:

```
A Microcosm plugin should be a function following the signature function(app, options, next) or an object with a register method with the same signature.
```

ltk: nailed it :nail_care:

PR #66, opened 2015-09-04
Lifecycle actions

nhunzaker:
This is an entirely internal change, but it paves the way for lifecycle actions. This commit adds lifecycle methods for `getInitialState`, `serialize`, and `deserialize`.

It will now be possible to do the following:

```javascript
import { getInitialState } from 'microcosm/lifecycle'
import { addPlanet } from 'actions/planets'

const Planets = {
  reset() {
    return []
  },

  add(planets, props)  {
    return planets.concat(props)
  },

  register() {
    return {
      [getInitialState] : Planets.reset,
      [addPlanet] : Planets.add
    }
  }
}
```

The old way of using getInitialState is supported as well, but the plan is to phase this out when we hit `v10.0.0`. The goal is to remove "special" methods, and implement a more transparent application lifecycle. 

Later down the line, we might consider:

```javascript
import { getInitialState } from 'microcosm/lifecycle'
import { addPlanet } from 'actions/planets'

function reset() {
  return []
}

function add(planets, props)
  return planets.concat(props)
}

export default function Planet() {
  return {
    [getInitialState] : reset,
    [addPlanet] : add
  }
}
```

On top of that, we should also consider memoizing register functions to cut down on object creation. Just a thought...


src/send.js
```diff
@@ -0,0 +1,20 @@
...
+  let pool = typeof store.register === 'function' ? store.register() : false
+
+  if (!pool || type in pool === false) {
+    return state
+  }
+
+  let handler = pool[type]
+
+  return typeof handler === 'function' ? handler.call(store, state, payload) : handler
+}
```
nhunzaker: Pretty branchy, but we'll cut it down when we can make breaking changes...

src/send.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Send
+ * Communicate an action to stores
+ */
+
+module.exports = function send (store, state, { payload, type }) {
+  if (type in store) {
+    return store[type].call(store, state, payload)
+  }
```
nhunzaker: This conditionally allows stores that implement the "special" methods to continue to work. In Microcosm 10.0.0 I will remove them (and we'll add a deprecation working on the 9.0.0 branch)

ltk:
I'm excited to see where this goes with lifecycle actions, but I do think the removal of "special" methods makes this marginally more difficult to use. Development of stores still require special knowledge, it's just now in the form of a require location and lifecycle method names. Seems fine though. :+1:

nhunzaker:
I'm torn on that too; the balance of magic and difficulty. We could definitely hold on to it, in either case I like the new internal mechanics better.

I'm also in no rush to make another breaking change. Let's see how both feel on future projects.

Issue #28, opened 2015-05-22
Lifecycle methods?


jbelcher:
Just wanted to open an issue to think through a some potential hooks. When an action resolves? Before/after a dispatch? When a microcosm resets or replaces?

nhunzaker:
Just to get some use cases in here:

1. Undo/Redo - travel along revisions to start. Useful for users to return to a previous state
2. Rewind/FastForward - travel along resolved actions. Useful for developers to debug complete life cycles

jbelcher:
Might also be a good way to do more with error states/rollbacks

nhunzaker:
Actually, I _almost_ put that. You could monitor actions, persist them in the background, and handle reconciling the returned record or rolling back if need be.

jbelcher:
Yeah FWIW that's super attractive to me—a lot of flux libraries get too opinionated w/ magic/abstractions. around things like that. I really like how clean/clear microcosm is in its current state.

nhunzaker:
Played around a bit with it today. Right now we call transactions "signals", but it's an internal API; I can change it without much fuss.

So far I have:

```javascript
didStart     : function(app) {},
willDispatch : function(app, transaction, action, result) {},
didDispatch  : function(app, transaction, action, result) {}
```

I have this in the https://github.com/vigetlabs/microcosm/compare/master...lifecycles branch with some interesting results:

![redo](https://cloud.githubusercontent.com/assets/590904/7799316/af55c8d0-02d2-11e5-9cc5-6114a45f82b4.gif)

jbelcher:
Really nice—feels flexible too. You can persist app states or signals and annotate either pretty easily w/ what's there.

nhunzaker:
Cool. There's no going back from this (a major release, I guess :smile_cat:) , so I definitely want to get this right. 

I'd like to continue to mature a reference implementation. I have a few unanswered questions:

1. How does undo/redo expose itself in an application?
2. I sort of stumbled upon this idea of treating actions transactionally, but I really like it. Should we ensure some sort of resolution order?
3. What about rewriting transaction history? [Datomic has the idea of retraction](http://docs.datomic.com/transactions.html#sec-3-6), but maybe that is too high level for Microcosm. Still, what if each transaction had an enabled state or something that allowed you to "forget" that something happened? 

I do not think that we need answers for questions 2 and 3 yet, but I am very curious about question 1.

nhunzaker:
[...]

PR #68, opened 2015-09-14
Transaction pool

nhunzaker:
This PR sets up an object pool for transactions and greatly simplifies the transaction creation process.

### DIFF
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -8,7 +8,6 @@ let send = require('./send')
 let tag = require('./tag')
 let eventually = require('./eventually')
 let sortById = require('./sortById')
-let has = require('./has')
 
 let Microcosm = function() {
   /**
@@ -50,7 +49,7 @@ Microcosm.prototype = {
     transaction.payload  = payload
     transaction.complete = complete
 
-    if (!complete && has(this.transactions, transaction) === false) {
+    if (this.transactions.indexOf(transaction) === -1) {
       this.transactions = this.transactions.concat(transaction).sort(sortById)
     }
   },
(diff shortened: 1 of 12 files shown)


src/Microcosm.js
```diff
@@ -39,68 +41,43 @@ let Microcosm = function() {
...
-    this.transactions.splice(this.transactions.indexOf(transaction), 1)
-  },
-
-  transactionDidFail(transaction) {
-    this.release(transaction)
-    return this.rollforward()
+    if (!complete && has(this.transactions, transaction) === false) {
+      this.transactions = this.transactions.concat(transaction).sort(sortById)
+    }
   },
```
nhunzaker: Actually, I always want to push these in to maintain history for those who want it.

src/Transaction.js
```diff
@@ -4,39 +4,29 @@
...
-  return {
-    type: `${ type }`,
-    payload,
-    meta: {
-      active: false,
-      done: false
-    },
-    error: false
-  }
+const blueprint = {
```
solomonhawk: Capitalize const bindings?

nhunzaker: Yes

src/lifecycle.js
```diff
@@ -4,13 +4,15 @@
  * actions.
  */
 
-exports.mapping = {
+const mapping = {
```
solomonhawk: Uppercase

solomonhawk:
This looks really nice! :+1: Can you give an example of using the transaction lifecycle hooks?

nhunzaker:
@solomonhawk The plan is to eventually expose those to plugins, but for stores, the lifecycle hooks are actions, so you can register to them like:


```
import { willStart } from 'microcosm/lifecycle'

let Store = {
    reset() {
        return []
    },
    register() {
        return {
            [willStart]: Store.reset
        }
    }    

}
```

Or, of course, you could just use `Store.getInitialState`.  The long term plan is to expose more hooks, but I wanted to get this transaction stuff clamped down first.

nhunzaker:
Basically, a method that will call the same hooks on Plugins. Soon to come! I believe they will be something like:

```
transactionDidOpen
transactionDidUpdate
transactionDidClose
transactionDidFail
```

nhunzaker:
(then I can finally close #28 )

PR #70, opened 2015-09-23
V8 optimized dispatch and refactor

nhunzaker:
Did some performance optimization to `send` and `dispatch` for the fun of it. In the process, I also cleaned up the code quite a bit. 

The moral of the story is "don't access shared memory in functions".


h0tl33t:
Neat!  :+1:

PR #72, opened 2015-09-25
Respect bound callbacks. Add greater test coverage

nhunzaker:
I want to support the following behavior:

```
app.push(action, params, reactComponent.method)
```

Up until now I was binding all `push` callbacks to `app`. That is problematic for this use case. Instead, I now use `call` inside of an anonymous function to respect `Function::bind`.

While I was at it, I also added test coverage to make sure that errors throw inside of `eventually` are not absorbed by promises. What a mind bender...

### DIFF
--- a/src/__tests__/eventually-test.js
+++ b/src/__tests__/eventually-test.js
@@ -1,12 +1,39 @@
+import Promise from 'promise'
 import assert from 'assert'
 import eventually from '../eventually'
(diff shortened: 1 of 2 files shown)


h0tl33t:
:+1:

solomonhawk:
This seems like a good change. :+1:

PR #73, opened 2015-09-28
Transaction tree

nhunzaker:
This PR marks the beginning of storing transactions as a directed graph instead of an array. Additionally, it exposes an API for controlling history.

undo-tree


src/Microcosm.js
```diff
@@ -71,31 +49,34 @@ Microcosm.prototype = {
...
-    }
+  clean(transaction, next) {
+    if (!transaction.complete) return
 
-    return this
+    this.base = dispatch(this.stores, this.base, transaction)
+    this.history.remove(transaction)
+
+    next()
   },
```
nhunzaker: This is the default merger strategy. It means: "progress through the current transaction pathway, merging transactions that are complete. Stop when you it an incomplete transaction"

src/__tests__/Graph-test.js
```diff
@@ -0,0 +1,207 @@
...
+
+    graph.connect('one', 'two')
+    graph.connect('two', 'three')
+    graph.connect('one', 'three')
+
+    assert.equal(graph.after('one'), 'three')
+    assert.equal(graph.before('three'), 'one')
+  })
+
+})
```
nhunzaker: Phew :sweat:

src/lifecycle.js
```diff
@@ -7,7 +7,8 @@
 const MAPPING = {
   'willStart'       : 'getInitialState',
   'willSerialize'   : 'serialize',
-  'willDeserialize' : 'deserialize'
+  'willDeserialize' : 'deserialize',
+  'willReset'       : '__AppWillReset'
```
nhunzaker: I have decided not to expose this for now. Any ideas on a good lifecycle method name?

Makefile
```diff
@@ -22,7 +22,7 @@ docs: $(DIST)
 	cp -r $@ $^
 
 javascript: $(DIST)
-	@babel -d $^ $(JS)
+	@babel --plugins babel-plugin-unassert -d $^ $(JS)
```
nhunzaker: Super handy. I'm using `assert` to unit test hard to reach cases, stripping it out for production builds.

src/__tests__/Graph-test.js
```diff
@@ -0,0 +1,207 @@
...
+    assert.equal(graph.focus, 'two')
+  })
+
+  it ('removing a focused node sets focus to false if there is no next node', function() {
+    let graph = new Graph()
+
+    graph.add('one')
+    graph.remove('one')
+
+    assert.equal(graph.focus, false)
```
efatsi: is `false` really the proper value for this situation?

efatsi: vs. `null`

src/__tests__/Graph-test.js
```diff
@@ -0,0 +1,207 @@
...
+
+    graph.append('one')
+    graph.append('two')
+    graph.append('three')
+    graph.setFocus('two')
+    graph.append('four')
+    graph.append('five')
+
+    assert.deepEqual(graph.pathway(), [ 'one', 'two', 'four', 'five' ])
+  })
```
efatsi: this test looks very similar to the one above it

src/__tests__/Graph-test.js
nhunzaker: :+1:. That is a really good idea... `focus`, `after`, and `before` should be set or return `null`, not `false`

nhunzaker: Done

src/__tests__/Graph-test.js
nhunzaker: Good call. Removing first test

nhunzaker:
@efatsi If you care to take a gander again. I realized that there is no reason to keep track of a `nodes` list in `Graph`. Instead, I just keep track of edges. Also, I added a `tree` method to make it easier to generate output.

nhunzaker:
Oh yeah, and it's muuuuch faster:

```
~/Development/microcosm transaction-tree
●° node bench/pathway.js
Creating a single path at 1000000 length
Appending: 346.94ms
Pathway: 107.792ms
```

bench/pathway.js
```diff
@@ -0,0 +1,21 @@
...
+}
+var after = time.now()
+
+console.log('Appending:', (after - start) / 1000 + 'ms')
+
+var beginPath = time.now()
+var items = graph.pathway()
+var endPath = time.now()
+
+console.log('Pathway:', (endPath - beginPath) / 1000 + 'ms')
```
nhunzaker: Hopefully we will never have a single path at 1 million nodes. The benchmark at ten thousand nodes is still impressive:

```
●° node bench/pathway.js
Creating a single path at 10000 length
Appending: 2.67ms
Pathway: 1.315ms
```

ltk: I didn't get what `bench` meant at first. Is this a common name, or would `benchmarks` be better?

src/Graph.js
```diff
@@ -0,0 +1,110 @@
+/**
+ * Graph
+ * A directed graph. Used to maintain a signal graph of transactions
+ * over time.
+ */
+
+let assert = require('assert')
```
ltk: doesnt look like you're using this in here

src/Microcosm.js
ltk: so things are only undoable before you commit the transaction?

src/lifecycle.js
ltk: `willReset` seems pretty good to me

ltk:
This is really neat. two questions:

- What are the memory implications of this change?
- Do you have a sense of real use case where one would want to switch between change branches on the history tree?

bench/pathway.js
nhunzaker: I'll rename it benchmarks

src/Graph.js
nhunzaker: Not any more. Good call, :+1:

nhunzaker:
@ltk 

_**What are the memory implications of this change**_

Before this change, we stored transactions in an array. Now they are kept in an array of two element arrays. The default cleanup strategy is to merge all completed transactions. That is to say, all actions that return primitive values, resolved promises, and fully iterated generators. 

This graph will not create an edge unless there are two transactions to join together. Using the default cleanup strategy, edges are never created and the penalty is never encountered

For cleanup strategies that support history, it'll assign an array for each edge. However it only stores the _transaction_, not the final state. This means that space is limited to the payload of the action. Framework memory usage should be rather low, the bottleneck would really be a matter of the size of the action payloads. 

I could see a future where we have finer grained undo, where we can merge together complete transactions in the middle of history. That would allow for even more powerful undo; we just aren't there yet.

So, TL;DR; I don't think it's an issue. Certainly for existing apps, there should be no effect at all. For other use cases, I am curious if there is a more optimal way to do it. However I think it's pretty good now, and the only other alternative I can think of would be to redefine the `Transaction` class as a node in a trie. This implementation keeps them separate, which I like.

_**Do you have a sense of real use case where one would want to switch between change branches on the history tree?**_

The specific use case I am thinking of is the ability to compare two on going changes side by side in Colonel Kurtz. Let's say you have a page, but you want to change the hero graphic. You aren't sure which one you like better, so you toggle between the two trees to preview what they look like (this is assuming we add live preview eventually)

src/lifecycle.js
nhunzaker: Sure. Let's expose it. I'll be sure to write up a pretty clear warning in the changelog

ltk:
Nice :+1: all the way.

src/Microcosm.js
nhunzaker: This is actually the clean up step. The default behavior is to merge down actions that are completely resolved. For example:

```
app.push(actionThatReturnsAPrimitive)
// => resolved immediately
// => merged
// => emit change event
```

```
app.push(actionThatReturnsAPromise)
// => wait to resolve
// => promise resolves
// => merged
// => emit change event
```

```
app.push(actionThatReturnsAPromise)
app.push(actionThatReturnsAPrimitive)
// push promise action into transaction graph
// push primitive action into transaction graph after promise
// => dispatch primitive
// => promise resolves
// => dispatch promise, primitive
// => merge promise, primitive
// => emit change
// => no outstanding transactions
```

nhunzaker:
Well I just bumped into an issue. I should have thought of this earlier, but restructuring the graph as a tree for visualization bumps into maximum call stack territory on deep chains. Unfortunately we can't lean on tail call optimization either.

This should probably be envisioned as a tree.

benchmarks/pathway.js
```diff
@@ -0,0 +1,21 @@
+var Graph  = require('../dist/src/Graph')
```
averyvery: Do you already have a script or command that runs these benchmarks, or are they not for public consumption?

averyvery:
Someone else should look at this, but :+1:

benchmarks/pathway.js
nhunzaker: Not for public consumption right now. Actually in a future branch I pull in `babel/register` to make the build process unnecessary.

nhunzaker:
Thanks, everyone. I'm going to close this out. Tree recursion issues is a complete show stopper. I've got an outstanding branch that eliminates this by maintaining a tree structure on insertion (instead of now where it's determined via recursion as a public API).

More to come! Benchmarks are already super promising:

```
●° node --expose-gc benchmarks/pathway.js
Working.............
┌──────────┬─────────┬──────────┬───────┬───────┬────────┬────────┬───────────────┐
│ Branches │ Nodes   │ Append   │ Path  │ Root  │ Values │ Prune* │ Memory Growth │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 1        │ 100,000 │ 0.0004ms │ 6.6ms │ 3.2ms │ 6.8ms  │ 22.6ms │ 0.29%         │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 2        │ 50,000  │ 0.0007ms │ 4.0ms │ 1.2ms │ 4.2ms  │ 19.1ms │ -9.07%        │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 3        │ 33,333  │ 0.0011ms │ 5.3ms │ 1.2ms │ 5.2ms  │ 20.0ms │ 0.18%         │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 4        │ 25,000  │ 0.0014ms │ 4.4ms │ 1.1ms │ 4.6ms  │ 20.3ms │ 0.19%         │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 5        │ 20,000  │ 0.0017ms │ 5.3ms │ 1.1ms │ 4.5ms  │ 23.4ms │ 0.17%         │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 6        │ 16,666  │ 0.0022ms │ 5.5ms │ 1.5ms │ 4.5ms  │ 22.4ms │ 0.17%         │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 7        │ 14,285  │ 0.0026ms │ 4.4ms │ 1.1ms │ 5.4ms  │ 23.0ms │ 0.17%         │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 8        │ 12,500  │ 0.0029ms │ 5.3ms │ 1.1ms │ 6.5ms  │ 21.7ms │ 0.17%         │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 9        │ 11,111  │ 0.0034ms │ 4.4ms │ 1.3ms │ 4.5ms  │ 20.3ms │ 0.17%         │
├──────────┼─────────┼──────────┼───────┼───────┼────────┼────────┼───────────────┤
│ 10       │ 10,000  │ 0.0038ms │ 5.7ms │ 1.2ms │ 4.5ms  │ 20.2ms │ 0.17%         │
└──────────┴─────────┴──────────┴───────┴───────┴────────┴────────┴───────────────┘
 * for entire branch
```

PR #74, opened 2015-09-29
Create `shouldTransactionMerge` method. Fix chatbot example

nhunzaker:
Walked through some things with @efatsi. This is the resulting PR.


src/Microcosm.js
```diff
@@ -55,11 +55,17 @@ Microcosm.prototype = {
     return this
   },
 
+  shouldTransactionMerge() {
+    return true
+  },
```
nhunzaker: Default merger strategy. This should allow users to interact with undo history at a much higher level of abstraction. 

For example, if you wanted to only use the last 500 changes, you could do something like:

```
shouldTransactionMerge() {
    return this.history.path().length > 500
}
```

Based on some performance metrics, this will execute in `0.49ms`, which is fast enough for me.

efatsi: just a note that your example (I believe) will just save the first 500 transactions, and then merge the rest in. As apposed to saving the last 500 (which would probably involve merging the early transactions into each other).

efatsi:
:+1:

src/Microcosm.js
nhunzaker: This will save the last 500 transactions, on the 501st, the first transaction will be merged in; working from left to right.

Actually, I will add test coverage for specifically this example.

nhunzaker: What are your thoughts on this test here? 2ed48d3

averyvery:
:+1:

PR #75, opened 2015-10-01
Allow generators to use generators

nhunzaker:
This PR allows generators actions to compose other generator actions. Really, it fixes a bug where yielding generators returns the generator itself to the store. Keeping with the expectation, i want to continue to break down the complex data types until I get a primitive value.

### DIFF
--- a/src/__tests__/dispatch-generator-test.js
+++ b/src/__tests__/dispatch-generator-test.js
@@ -99,4 +99,35 @@ describe('When dispatching generators', function() {
 
     app.push(single, 4)
   })
+
+  it ('runs results through coroutine', function(done) {
+    let app = new Microcosm()
+
+    function* range(start, end) {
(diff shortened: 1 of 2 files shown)


nhunzaker:
Just to give some context here.

This came up queuing animation states for a project. I want to compose them together in interesting ways. With this PR fixes a bug I found doing:

```
function* stageIn() {
    yield new KeyFrame({ state: 'intro'' }, 0)    
    yield new KeyFrame({ state: 'intro-leave' }, 200)
}

function* stageOut() {
    yield new KeyFrame({ state: 'curtain'' }, 400)    
}

function* transition(page) {
    yield stageIn()
    yield new KeyFrame({ state: 'page' })
    yield stageOut()
}
```

Purely a thought experiment, but it did uncover the bug :)

PR #76, opened 2015-10-06
Transaction tree tweaks

nhunzaker:
Second attempt...

This PR moves away from a graph like data structure and implements history as a tree. It's also incredibly fast:

```
Nodes  ::append()  ::root()  ::reduce(merge)  ::size()  ::prune()  Memory Growth
-----  ----------  --------  ---------------  --------  ---------  -------------
10000  0.0008ms    0.40ms    1.81ms           0.32ms    1.24ms     0.883%

Dispatching 10000 actions: 14ms
```

Performance is Similar in Safari and slightly less stellar in Firefox. In any case: 10,000 unique transactions processed is a very lofty goal. This is essentially pushing a new action ever 16 milliseconds for 2.6 minutes. If we ever want to enable high fidelity rewind, we can definitely do it...

undo-tree


examples/undo-tree/index.jsx
```diff
@@ -0,0 +1,53 @@
+import Microcosm from 'Microcosm'
+import React from 'react'
+import Drawing from './components/drawing'
+
+class UndoTree extends Microcosm {
+
+  shouldHistoryKeep(transaction) {
+    return false
+  }
```
nhunzaker: Thoughts on this name, @efatsi?

examples/undo-tree/index.jsx
```diff
@@ -0,0 +1,53 @@
+import Microcosm from 'Microcosm'
+import React from 'react'
+import Drawing from './components/drawing'
+
+class UndoTree extends Microcosm {
+
+  shouldHistoryKeep(transaction) {
+    return true
+  }
```
nhunzaker: Sorry to repost... bad push. @efatsi thoughts?

efatsi: sounds like a good name to me. whatever the name though, this logic is not trivial and deserves a good writeup in the readme i believe.

nhunzaker: I agree.

nhunzaker: Really, I continue to struggle for a name here. It's "should the app commit this transaction into base state and purge it from history"

efatsi: you could always go the verbose route `shouldMergeTransactionIntoBaseStateAndPurgeFromHistory`

nhunzaker: Haha, you actually beat me to typing that exact thing out!

PR #78, opened 2015-10-16
Key paths

nhunzaker:
When upgrading Colonel Kurtz to the latest version of Microcosm, I bumped into some friction moving away from our old API that allowed:

```
app.set('abitrary-key', stuff)
```

This allowed a user to directly modify application state, which was extremely problematic when we moved to storing changes as a series of transactions over time. I'm really glad we moved away from this, however it did expose an issue: 

All stores must be mounted to a namespaced key, so there's no efficient way to directly assign a single-value key to the root state object. This makes it cumbersome to perform operations on the entire state object itself.

This PR adds support for this by implementing key path access to app state. This means that a store can be mounted to an array of keys and Microcosm will drill into the global app state object to assign a new value. 

All modifications occur immutably, which means we take about a 3-4ms penalty on our dispatch benchmark. That isn't great, but I think I can make up the efficiencies somewhere else in the pipeline. Regardless of the performance regression (which is over 10,000 transactions, more than 200 times what I've ever seen in the wild), I've bumped into this several times working on projects with Microcosm, and I think it's an important addition.

```javascript
import { willStart } from 'microcosm/lifecycle'

app.addStore(function() {
    return {
        [willStart]: { test: true }
    }
})

assert(app.state.test)
```

```javascript
app.addStore([ 'foo', 'bar', 'baz' ], {
    getInitialState() {
        return true
    }
})
assert(app.state.foo.bar.baz)
```

### DIFF
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -51,11 +51,9 @@ Microcosm.prototype = {
    * a new state. This is the state exposed to the outside world.
    */
   rollforward() {
-    let next = merge({}, this.base)
-
-    this.state = this.history.branch().reduce((state, transaction) => {
+    this.state = this.history.branch().reduce((state, transaction, i) => {
       return dispatch(this.stores, state, transaction)
-    }, next)
+    }, this.base)
 
     this.emit(this.state)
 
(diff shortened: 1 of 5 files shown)


ltk:
This looks great :+1:

PR #84, opened 2015-11-18
Nh revisit start and tree

nhunzaker:
@solomonhawk and @mzlock found some fun issues in transaction history that would cause store handlers to execute more times than expected. We've addressed that in the latest beta release (9.14.1-beta), however I wanted to rethink this a bit.

There are a couple of changes:

- Do not re-evaluate state when a store is added. In future releases, I may consider throwing an error if a store is added after a Microcosm starts.
- `:start()` pushes the `willStart` lifecycle action instead of resetting. This sounds pretty obvious, I'm not sure why it wasn't this way before.
- Removed some `null` optimization checks that aren't necessary.

### DIFF
--- a/benchmarks/dispatch-performance.js
+++ b/benchmarks/dispatch-performance.js
@@ -52,6 +52,8 @@ app.addStore('three', Store)
 app.addStore('four',  Store)
 app.addStore('five',  Store)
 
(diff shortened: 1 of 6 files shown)


ltk:
:+1:

PR #87, opened 2015-11-30
Tree should push children to beginning of list. Small Tree.js refactoring.

nhunzaker:
This commit changes the order in which child nodes of the Tree data structure are appended. This allows for "redo" functionality that moves forward to the most recent change.

undo


h0tl33t:
Neat! :+1:

PR #88, opened 2015-11-30
Add push benchmark

nhunzaker:
This PR adds another benchmark for general push performance. It's near instant, but I wanted to set up a benchmark before exploring some potential updates to this function.

screen shot 2015-11-30 at 9 16 09 am

### DIFF
--- a/benchmarks/push-performance.js
+++ b/benchmarks/push-performance.js
@@ -16,7 +16,7 @@ action.toString = function () { return 'test' }
 var Store = function() {
   return {
     getInitialState: 0,
-    test: function(n) { return n + 1}
+    test: function(n) { return n + 1 }
   }
 }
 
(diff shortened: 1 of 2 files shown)


benchmarks/push-performance.js
```diff
@@ -0,0 +1,52 @@
...
+
+var app = new Microcosm()
+
+var action = function test () {}
+action.toString = function () { return 'test' }
+
+var Store = function() {
+  return {
+    getInitialState: 0,
+    test: function(n) { return n + 1}
```
h0tl33t: Spacing :tm: at the end of the return statement.

h0tl33t:
Such fast.  Very speed.  :+1:

benchmarks/push-performance.js
nhunzaker: Ahhhh!

PR #89, opened 2016-01-08
Breakout history

nhunzaker:
This PR extracts the history API into a separate module. Ideally, this could eventually be overridden. It also lays the groundwork for a consistent set of lifecycle methods for transactions.


src/histories/Branches.js
```diff
@@ -0,0 +1,63 @@
...
+ * Branches History
+ * Stores transaction history as a tree.
+ */
+
+import Tree from '../Tree'
+
+let Branches = function (options={}) {
+  this.base = {}
+  this.tree = new Tree()
+  this.maxHistory = options.maxHistory || -Infinity
```
averyvery: why the -?

averyvery:
:+1: really sharp

src/histories/Branches.js
nhunzaker: By default, I never want to keep any history. This feels more permanent than `0`.

Also I never miss a chance to use Infinity :)

nhunzaker:
Giving my own :+1:  to circumvent our slackbot :P

PR #90, opened 2016-01-13
Plugins can be synchronous

nhunzaker:
If a plugin does not provide a `next` argument, it would be considered synchronous. This is to prevent the tedium of calling `next()` for otherwise synchronous tasks:

```javascript
const Logger  = {

  register(app, options) {

    app.listen(function() {
      console.log("I changed!")
    })
}
```

### DIFF
--- a/test/plugin-test.js
+++ b/test/plugin-test.js
@@ -112,4 +112,14 @@ describe('Plugins', function() {
 
     app.start()
   })
+
+  it ('just calls through plugins that have no next callback', function(done) {
+    let app = new Microcosm()
+
+    app.addPlugin(function(app, _) {
+      // intentionally nothing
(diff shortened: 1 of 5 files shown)


test/plugin-test.js
```diff
@@ -112,4 +112,14 @@ describe('Plugins', function() {
 
     app.start()
   })
+
+  it ('just calls through plugins that have no next callback', function(done) {
+    let app = new Microcosm()
+
+    app.addPlugin(function(app, _) {
+      // intentionally nothing
```
nhunzaker: This doesn't test that the plugin was actually called. Fixing...

solomonhawk:
This is a good change. :+1:

PR #91, opened 2016-01-14
Move bad action pushing to app. Improve message

nhunzaker:
I hit this on a project and it got me thinking... Ideally the `Transaction` module shouldn't be presented to the developer. This is a private module. With that in mind, I moved the "bad action" error to `app.push`, and improved the messaging:

screen shot 2016-01-14 at 9 06 28 am


cwmanning:
:+1:

greypants:
:+1:

PR #94, opened 2016-01-26
Lifecycle plugins

nhunzaker:
This PR breaks out all internal logic for Microcosm into a plugin architecture with lifecycle methods. Certain key lifecycle methods are extendable via plugins.

Specifically, there are 4 new lifecycle methods:

1. `willUpdate` Executed to determine new state.
2. `willAddStore` Executed right before a store is created.
3. `willAddPlugin` Executed right before a plugin is created.
4.  `willOpenTransaction` Executed right before a transaction is created

Additionally, the following lifecycle methods are available to plugins for similar purposes:

1. `serialize`
2. `deserialize`
3. `getInitialState`

**293 additions and 327 deletions. Nice!**


examples/undo-tree/client.js
```diff
@@ -1,8 +1,10 @@
 import Render   from './app/plugins/render'
 import UndoTree from './app/undo-tree'
+import Debugger from 'microcosm-debugger'
```
greypants: oh?

nhunzaker: Baaaaby steps, but it didn't feel like it was necessary anymore. Check it out:

https://github.com/vigetlabs/microcosm-debugger

Help me make it better!

nhunzaker: Specially now that we have 7 new lifecycle methods

src/Microcosm.js
```diff
@@ -1,18 +1,17 @@
...
-import dispatch    from './dispatch'
-import eventually  from './eventually'
 import flatten     from './flatten'
 import install     from './install'
 import lifecycle   from './lifecycle'
 import merge       from './merge'
 import tag         from './tag'
 
-let Microcosm = function() {
+function Microcosm (options) {
```
greypants: I think we should standardize on this. Named functions for the win.

nhunzaker: Seriously.

nhunzaker: Specifically, for constructors! I totally want the name Microcosm to show up in the console.

src/Microcosm.js
```diff
@@ -20,104 +19,50 @@ let Microcosm = function() {
...
-   * and the change will disappear from history.
-   */
   push(action, params, callback) {
-    if (process.env.NODE_ENV !== 'production' && action == null) {
-      throw new TypeError(`Unable to perform: app.push(${ action })\n\n`
-                          + 'This typically happens when an action is accessed from the wrong key of an object, such as:\n\n'
-                          + '• Actions.mispelledAction\n'
-                          + '• import { mispelledAction } from "actions"')
-    }
+    let transaction = new Transaction(tag(action))
```
greypants: What's the definition of a transaction?

nhunzaker: A transaction in microcosm is a distinct point in time. Every time a transaction is updated, it recalculates state. Application state is the result of dispatching the actions associated with each transaction to a Microcosm's stores. 

This lets you do really smart time travel and have complete predictability on execution order of actions:



Cool, yeah?

src/Transaction.js
```diff
@@ -3,12 +3,38 @@
...
+  action   : null,
+  active   : false,
+  error    : false,
+  payload  : undefined,
+  complete : false,
+
+  execute(params, onNext, onComplete, scope) {
+    let body = this.action.apply(null, flatten(params))
+
+    return coroutine(body, (error, payload, done) => {
```
greypants: > A coroutine is a function that can suspend its execution (yield) until the given given YieldInstruction finishes.
TYMK :stars:

src/Microcosm.js
nhunzaker: Sorry, more concretely. Whenever you push an action with `app.push`, it creates an associated transaction. This transaction _eventually_ resolves and dispatches the result to stores. This is how Microcosm "waits" for promises to finish resolving and how optimistic updates work.

src/Transaction.js
nhunzaker: Yes. Every time it yields a new value, state rolls forward with that new value. This is how we can do optimistic updates without having to clean up.

examples/undo-tree/app/undo-tree.js
```diff
@@ -4,27 +4,8 @@ import Pixels    from './stores/pixels'
 export default class UndoTree extends Microcosm {
 
   constructor() {
-    super()
+    super({ maxHistory: Infinity })
```
ltk: isn't `Infinity` the default `maxHistory` now?

src/Microcosm.js
greypants: very cool. How long do you hang onto those?

examples/undo-tree/server.js
```diff
@@ -1,6 +1,6 @@
 /**
  * This is the main server entry point into the
- * undo-tree example.
+ * drawing example.
```
ltk: but the directory name is still `undo-tree` :-(

greypants:
:+1: Is this all just internal stuff, or are there any user use-cases that could go in the README?

src/Microcosm.js
```diff
@@ -20,104 +19,50 @@ let Microcosm = function() {
...
-  transactionWillUpdate(transaction, error, payload, complete) {
-    transaction.active   = !error
-    transaction.complete = complete
-    transaction.error    = error
-    transaction.payload  = payload
-  },
-
-  transactionWillClose() {
-    this.history.prune(this.clean, this)
+  lifecycle(type, initial={}) {
```
ltk: thinking `stage`, or `eventType`, or `hook` might be a more descriptive name for `type`

src/Microcosm.js
```diff
@@ -20,104 +19,50 @@ let Microcosm = function() {
...
-
-      if (complete) {
-        this.transactionWillClose(transaction)
-        eventually(callback, this, error, payload)
-      }
-
-      this.rollforward()
-    })
+    return transaction.execute(params, this.rollforward, callback, this)
   },
```
ltk: this is a bit of a tricky signature

src/Microcosm.js
```diff
@@ -20,104 +19,50 @@ let Microcosm = function() {
...
-      this.transactionWillUpdate(transaction, error, payload, complete)
-
-      if (complete) {
-        this.transactionWillClose(transaction)
-        eventually(callback, this, error, payload)
-      }
-
-      this.rollforward()
-    })
+    return transaction.execute(params, this.rollforward, callback, this)
```
ltk: SubmissionUpvotes

src/Microcosm.js
```diff
@@ -179,23 +101,37 @@ Microcosm.prototype = {
...
+   * a given state.
+   */
+  reset(state, callback) {
+    return this.push(lifecycle.willReset,
+                     this.lifecycle(lifecycle.willReset, state),
+                     callback)
+  },
+
+  /**
+   * Resets to a given state, passing it through deserialize first
```
ltk: sry, grammar consistency nitpick :tm:: period at the end?

src/Microcosm.js
ltk: or `action`

src/Transaction.js
```diff
@@ -3,12 +3,38 @@
...
-    payload  : payload,
-    complete : complete
+import coroutine  from './coroutine'
+import flatten    from './flatten'
+import eventually from './eventually'
+
+export default function Transaction (action, payload) {
+  this.action  = action
+  this.type    = `${ action }`
+  this.active  = arguments.length > 1
```
ltk: just curious: why not check for the definedness of `payload` instead?

src/Transaction.js
```diff
@@ -3,12 +3,38 @@
...
+  this.payload = payload
+}
+
+Transaction.prototype = {
+  action   : null,
+  active   : false,
+  error    : false,
+  payload  : undefined,
```
ltk: need this?

src/install.js
```diff
@@ -18,11 +18,14 @@ function ensureCallback (plugin) {
   // Otherwise wrap synchronous plugins to match the async flow
   return function (app, options, next) {
     plugin.register(app, options)
+
     next(null)
   }
 }
 
 function installPlugin (next, plugin) {
+  if (!plugin.register) return next
```
ltk: should this be considered an error state? is there a case where someone would want to add a plugin with no register function?

examples/undo-tree/app/undo-tree.js
nhunzaker: No. It's actually `-Infinity`. By default we only hold on all incomplete transactions. When they finish, we merge them into a cache (See the history plugin for more)

src/Microcosm.js
nhunzaker: Until they complete.

src/Microcosm.js
nhunzaker: Yes.... Eventually I want this to be:

```
transaction.execute(params, 
                    this.transactionDidUpdate, 
                    this.transactionDidFail,
                    this.transactionDidComplete)
```

But I didn't want to totally re-write co-routine at the same time.

nhunzaker: (These would be _additional_ lifecycle hooks too :))

examples/undo-tree/server.js
nhunzaker: Uh.. my bad. I'll rename this guy.

src/install.js
ltk: I see now. Just had to read further.

src/Microcosm.js
nhunzaker: I like hook.

examples/undo-tree/server.js
nhunzaker: Renamed to painter!

nhunzaker:
Eventually, this will all be documented in the readme (and plugins API
doc). That'll be the next step, and a next PR.

src/Microcosm.js
nhunzaker: Made `hook`

src/Transaction.js
nhunzaker: Yes. I want to keep the shape and type information for transactions
consistent for better V8 optimization. These these things will be operated
on pretty frequently for certain use cases.

src/install.js
nhunzaker: Yeah. It's for plugins that don't implement register, but want to use
another lifecycle hook.

src/Transaction.js
nhunzaker: The payload might be `undefined`, but we only use the second argument
internally to Microcosm. I'll give this a second look.

src/Microcosm.js
nhunzaker: Yes.

ltk:
This is terrific!

I do have some thoughts. (Sorry, they're are a little half-baked at the moment, but I wanted to get them out here for some discussion.)

I have a negative gut feeling about how plugins are tied to the microcosm lifecycle (relying on properly-named plugin functions).

Mainly:
- They're fragile - It's easy to mistype a function name and be baffled by the fact that your hook isn't executing. Also sadtimes for everybody if we want to rename a lifecycle action in the future.
- They're static - You can't do fancy dynamic things like you can do if you allowed anonymous functions to be executed in response to lifecycle actions.
- They're lonely - You can only have one per lifecycle action per plugin :-(
- They're (relatively) permanent - You can't easily deregister them in response to occurences in the app.

What do you think about a more formal registration process for lifecycle hooks, to which you can pass arbitrary and  anonymous functions, like so:

```js
import {lifecycle} from 'microcosm'

export default {
  register(app, options) {
    // Simplest example: can still reference a plugin function
    app.on(lifecycle.willRollForward, this.arbitrarilyNamedFunction)

    // Simple example: add an arbitrary function as a lifecycle hook
    app.on(lifecycle.willRollForward, (state) => {
      // Return a thing.
    })

    // Registration labels for more granular deregistration/ avoidance of duplicate registrations?
    app.on(lifecycle.willRollForward, this.arbitrarilyNamedFunction, 'some-label')
    app.off(lifecycle.willRollForward, 'some-label')

    // Complex example: dynamically adding/removing hooks
    app.listen((state) => {
      const trialTruncationLabel = 'trial-trunctation'

      if (state.user.trial) {
        app.on(lifecycle.willRollForward, (state) => {
          return truncateTestForFreeTrial(state)
        }, trialTruncationLabel)
      } else {
        // Deregistration
        app.off(lifecycle.willRollForward, trialTruncationLabel)
      }
    })
  },

  arbitrarilyNamedFunction(state) {
    // Return a thing.
  }
}
```

The obvious disadvantage of this approach is the extra complexity associated with storing lifecycle hooks apart from the plugin.

Another question would be: should/can a plugin know about all the hooks it has registered?

nhunzaker:
Fantastic thoughts! I'm a huge fan of lifecycle methods, sadly for a lot of the same reasons you've mentioned. Not every single one, but I thought I would work through your items directly:

> 1.They're fragile. [specifically around names]

_Yes_. We take a lot of measures when pushing actions and dispatching to stores to make sure we send out developer warnings when things are undefined or don't appear to be hooked up properly. I wonder how we can help to mitigate that with plugins.

As for names changing. Ideally, we never change these names (or bump a major release if we do). I'd like to get these names right the first time and would love some help crafting them. Are these lifecycle method names good? How could we improve them?

> 2. They're static

This is my favorite thing about lifecycle methods. The framework maintains control over a fast majority of the hooks. Execution order of lifecycle methods is easier to reason about and the scope of possible orderings is significantly limited.

If we decide to add an event based system, we need to critically think about how to prevent developers from getting into the same event subscription nightmares that plague Backbone apps.

> 3. They're lonely.

I know :(, but you can call multiple subroutines within the main lifecycle method, right? Also, plugins should be for a specific purpose. If a plugin is responding to a lifecycle event with more than one type of behavior -- should there be two plugins?

> 4. They're (relatively permanent)

We can't deregister plugins right now. Maybe we should add that. What would that look like?

---

Beyond that, Stores and Plugins are becoming incredibly similar. It may be worth doing a breaking release to make their API's consistent. We could evaluate an event based system at that time.

Thoughts?

nhunzaker:
Building from this. Should we consider a `10.0.0`  release where we rework the `register` function of Plugins to be identical to stores? Then Plugins could register to lifecycle events just like Stores:

```
import lifecycle from 'microcosm/lifecycle'

const Plugin = {
    doSomething(app, state) {
        // protected to lifecycle method string names changing.
    },

    register() {
        return {
            [lifecycle.willStart]: Plugin.doSomething
        }
    }
}

PR #95, opened 2016-01-27
Add connect plugin. Update React-Router example.

nhunzaker:
This PR adds a Provider and Connect addon similar to the `react-redux` project. Additionally, it updates the ReactRouter example to utilize these addons.

Connect will listen to app and prop changes to calculate new state. This state is injected into a child component as props. By default, this is pure. For projects that mutate data (please don't do this), you can pass an optional `{ pure: false }` to Connect.

This is inspired heavily by a combination of the NuclearJS and Redux implementations of this concept:

- https://github.com/jordangarcia/nuclear-js-react-addons/blob/master/src/connect.js
- https://github.com/rackt/react-redux/blob/master/src/components/connect.js

Basic usage:

```javascript
const Presenter = Connect(function (props) {
  let { id } = props.params

  return {
    list  : Lists.get(id),
    items : Items.childrenOf(id)
  }
})

const ListPage = Presenter(React.createClass({
    /// this component will have app, list, and items props
})

ReactDOM.render(<Provider app={ app }><ListPage /></Provider>)
```

Further, this means that the application instance does not need to be continually passed down to child components. The React `context` API bridges the gap for us. Connected components can accept an `app` property if not relying on context. This should make it easy to patch in an application when testing.


cwmanning:
:+1: nice job

examples/react-router/app/resources/lists/index.js
```diff
@@ -0,0 +1,4 @@
+export ListStore from './store'
+export * from './actions'
+
+export default from './getters'
```
greypants: I ended up killing the index file in my implementation... the extra boilerplate just to do fancy one-line `imports` didn't seem necessary. It's pretty though.

examples/react-router/app/components/lists/show.jsx
```diff
@@ -0,0 +1,61 @@
+import Connect from '../../../../../src/addons/connect'
+import ItemForm from './parts/item-form'
+import Items, { addItem, removeItem } from '../../resources/items'
```
greypants: it was unclear at first that `Items` is the microcosm store and `addItem, removeItem` are actions. I think I'd favor:

```
import Items from '../../resources/items/store'
import { addItem, removeItem } from '../../resources/items/actions'
```

See note below on the index file.

examples/react-router/app/resources/lists/index.js
nhunzaker: Yeah. I still I like the ergonomics of:

```
import Lists, { createList, removeList } from 'app/resources/lists'
```

nhunzaker: I wonder how we can get the best of both.

examples/react-router/app/components/lists/show.jsx
nhunzaker: To me, it's clear because of the nomenclature: `add*`, `remove*`

nhunzaker: Regardless, I'm cool just keeping these as separate imports for now.

test/addons/connect-test.jsx
```diff
@@ -0,0 +1,249 @@
...
+    let Namer = Connect(function(props) {
+      return { name: state => props.prefix + ' ' + state.name }
+    })(function({ name }) {
+      return (<p>{ name }</p>)
+    })
+
+    let component = Test.renderIntoDocument(<Namer app={ app } prefix="Colonel" />)
+
+    assertText(component, 'Colonel Kurtz')
+  })
```
greypants: is this testing something different than the test on line 15?

test/addons/connect-test.jsx
```diff
@@ -0,0 +1,249 @@
...
+    let Namer = Connect(function(props) {
+      return { name: state => props.prefix + ' ' + state.name}
+    })(function({ name }) {
+      return (<p>{ name }</p>)
+    })
+
+    let component = Test.renderIntoDocument(<Provider app={ app }>{ <Namer prefix="Sir" /> }</Provider>)
+
+    assertText(component, 'Sir Billy')
+  })
```
greypants: Where is this testing that `app` exists as a prop?

test/addons/connect-test.jsx
```diff
@@ -0,0 +1,249 @@
...
+      },
+      render() {
+        return this.state.display ? <Child app={ app } /> : null
+      }
+    })
+
+    let component = Test.renderIntoDocument(<Parent />)
+
+    component.setState({ display: false })
+  })
```
greypants: nice test.

test/addons/connect-test.jsx
nhunzaker: Very good question.... Excellent catch.

test/addons/connect-test.jsx
```diff
@@ -0,0 +1,249 @@
...
+
+      let Namer = Connect(function() {
+        return { name: state => state.name }
+      })(function({ name }) {
+        renders += 1
+        return (<p>{ name }</p>)
+      })
+
+      Test.renderIntoDocument(<Namer app={ app } />)
+      app.publish()
```
greypants: what's publish do? Just trigger an update?

test/addons/connect-test.jsx
nhunzaker: Take a look at 0674066

test/addons/connect-test.jsx
nhunzaker: It does now, yes.

nhunzaker: (Thanks, btw)

test/addons/connect-test.jsx
nhunzaker: Sounds good to me.

nhunzaker: See 1ad630a

examples/react-router/app/components/lists/show.jsx
greypants: Yeah, it's clear if you think about it... but you have to think. People looking at the example are coming in with no clue what they're doing, and would benefit from things being a little spelled out.

greypants:
:+1: that was fast!

examples/react-router/app/components/lists/show.jsx
nhunzaker: Check out cfc74db

PR #97, opened 2016-02-02
Further optimizations to tree

nhunzaker:
I realized a couple of things about the Tree structure while I was in History.js. This PR implements those ideas. Basically, everything is faster:

Before:

```
Nodes  ::append()  ::root()  ::reduce(merge)  ::size()  ::prune()  
-----  ----------  --------  ---------------  --------  ---------
10000  0.0008ms    0.42ms    2.57ms           0.35ms    1.41ms 
```

After:

```
Nodes  ::append()  ::root()  ::reduce(merge)  ::size()  ::prune() 
-----  ----------  --------  ---------------  --------  --------- 
10000  0.0003ms    0.0010ms  2.33ms           0.22ms    0.54ms
```

A lot of this boils down to assigning a `root` node. This makes calculating the `root` essentially free and significantly reduces the cost of pruning because it is not penalized calculating the root. 

Additionally, I added an optimization where a node's children list is only created when a child is added. This tremendously reduces memory usage for high frequency dispatches that immediately resolve (like listening to window resize.) It also makes append ~300% faster (which is near instant anyway....)

There's even less code, when you look at the specific changes to `./src`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,5 +65,5 @@ test-watch:
 
 bench: javascript
 	@ node --expose-gc --trace-deopt benchmarks/tree-performance
-	@ node --expose-gc --trace-deopt benchmarks/dispatch-performance
-	@ node --expose-gc --trace-deopt benchmarks/push-performance
+	@ node --trace-deopt benchmarks/dispatch-performance
+	@ node --trace-deopt benchmarks/push-performance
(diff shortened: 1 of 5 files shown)


cwmanning:
:+1:

PR #98, opened 2016-02-03
Start throws errors by default if not provided a callback

nhunzaker:
If `app.start()` is invoked and no callback is provided, errors passed through the plugin process will raise an exception. 

Basically, I'm tired of having to do this by hand on every single project:

```javascript
app.start(function(error) {
    if (error) {
        throw error
    }
})
```

### DIFF
--- a/test/Microcosm-test.js
+++ b/test/Microcosm-test.js
@@ -106,16 +106,16 @@ describe('Microcosm', function() {
     })
   })
 
-  it ('throws errors by default if a callback is not provided to start', function() {
-    let app = new Microcosm()
+  context('when a microcosm plugin passes an error', function() {
+    let app   = new Microcosm()
     let error = 'This error should exist!'
 
-    app.addPlugin(function willPassError (app, options, next) {
+    app.addPlugin(function willFail (app, options, next) {
       next(error)
     })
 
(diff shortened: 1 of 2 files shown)


test/Microcosm-test.js
```diff
@@ -106,4 +106,16 @@ describe('Microcosm', function() {
...
+    let app = new Microcosm()
+    let error = 'This error should exist!'
+
+    app.addPlugin(function willPassError (app, options, next) {
+      next(error)
+    })
+
+    assert.throws(function() {
+      app.start()
+    }, error)
```
mackermedia: I'm not sure I follow this test. Why do you need to add a plugin just to test that `app.start()` w/out a callback will throw an error?

nhunzaker: Maybe a bad test name. I want to test that plugins which call `next()` with an error that aren't caught in `app.start()` are eventually thrown.

nhunzaker: App.start installs all plugins, and they can fail.

mackermedia: Ah okay, cool.

mackermedia:
:+1:

PR #99, opened 2016-02-05
Unwrap promises when dispatching within Microcosm

nhunzaker:
This PR creates an escape hatch when Microcosm must respond to the callback of a promise. For internal operations, such as dispatching to stores, Microcosm is completely done with the promise. Any exceptions that are raised... should be raised.

Basically, I want to avoid this situation:

(try it here: https://tonicdev.com/56b4ca0eec3a740c00328daf/56b4ca0eec3a740c00328db0)
```javascript
var Microcosm = require('microcosm@9.19.0')
var Promise   = require('promise@7.1.0')

var app = new Microcosm()

function getUsers (userId) {
  return Promise.resolve({ name: 'Phil' })
}

app.addStore({
  register() {
    return {
      [getUsers]: () => { throw 'Oh snap!' }
    }
  }
})

app.start()

app.push(getUsers)        // where'd my error go?
app.push(getUsers).done() // ah there we go
```

Technically, this isn't really Microcosm's problem, but it's simple enough to implement and a prevents painfully silent bugs with promises; a tremendously better developer experience.


benjtinsley:
:+1:

src/unpromise.js
```diff
@@ -0,0 +1,49 @@
...
+ * of the JavaScript runtime loop.
+ *
+ * This has no effect on callbacks outside of Microcosm, since the
+ * chain of callbacks is never returned.
+ *
+ * @param {Promise} promise - The promise to escape out of.
+ * @param {Function} callback - The error-first callback to execute.
+ *
+ * @returns undefined
+ */
```
greypants: Nice documentation!

src/unpromise.js
```diff
@@ -0,0 +1,49 @@
...
+
+import eventually from './eventually'
+
+export default function liftCallback (promise, callback) {
```
greypants: why call the file `unpromise` and the function `liftCallback`?

nhunzaker: Because I renamed the file and didn't do project-replace :(. Good catch

mzlock:
:+1:

PR #100, opened 2016-02-10
Promise callbacks should receive action payload

nhunzaker:
This commit fixes a regression introduced in `9.6.0` where the promise payload was not being returned from `push`, this prevented promise chains from receiving the transaction payload.

Basically:

```javascript
function action (params) {
  return Promise.resolve(params)
}

app.push(action, 'foobar').done(function (payload) {
    assert.equal(payload, 'foobar')
})
```
Pretty safe bug fix. This does not effect action callbacks, like:

```javascript
app.push(action, params, function (error, payload) {
    // still correct behavior
})
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 9.19.1
+
+- Fixes a regression introduced in 9.6.0 where the promise payload was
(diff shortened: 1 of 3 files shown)


greypants:
:+1:

PR #103, opened 2016-02-16
Eliminate the need to set a timeout on every callback

nhunzaker:
Since the addition of the [unpromise](https://github.com/vigetlabs/microcosm/blob/master/src/unpromise.js) function, Promise try/catch behavior no longer affects Microcosm internals. This means that we do not need to execute the callback of `app.push` within a set timeout.

This PR makes that change, and makes some updates to a few timeout-related tests that cut the test suite by a couple hundred milliseconds.

Technically, this also probably means that app.push callbacks execute quite a bit faster, but the real value here is that there's simply less machinery.

### DIFF
--- a/src/Transaction.js
+++ b/src/Transaction.js
@@ -3,9 +3,8 @@
  * An account of what has happened.
  */
 
-import coroutine  from './coroutine'
-import flatten    from './flatten'
-import eventually from './eventually'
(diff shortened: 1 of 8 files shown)


greypants:
:+1:

mackermedia:
:+1:

PR #104, opened 2016-02-16
Do not clone objects when the target value is the same

nhunzaker:
Microcosm maintains a pair of methods for writing to `app.state`: `get` and `set`. This PR makes a change to `set` so that it first checks to see if a value has actually changed before overwriting it.


greypants:
:+1:

mackermedia:
:+1:

PR #107, opened 2016-02-18
Throw error if push() called before start()

greypants:
Adds a helpful error to let you know you probably forgot to call `app.start()` if trying to push before doing so.

### DIFF
--- a/src/Microcosm.js
+++ b/src/Microcosm.js
@@ -11,6 +11,12 @@ import lifecycle   from './lifecycle'
 import merge       from './merge'
 import tag         from './tag'
 
+function throwIfError (error) {
+  if (error) {
+    throw error
+  }
+}
+
 /**
  * Microcosm
  * @class
@@ -42,6 +48,7 @@ function Microcosm (options) {
 
 Microcosm.prototype = {
   constructor: Microcosm,
+  started: false,
 
   /**
    * Calculates initial state. This is no longer used internally,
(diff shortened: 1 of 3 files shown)


src/Microcosm.js
```diff
@@ -226,7 +234,8 @@ Microcosm.prototype = {
    * @param {function} [callback] - A callback to execute after the application starts
    * @return {Microcosm} The invoking instance of Microcosm
    */
-  start(callback) {
+  start(callback=throwIfError) {
+    this.started = true
     install(this.plugins, callback)
     return this
   }
```
greypants: @nhunzaker If this is anywhere but on the first line, it breaks all the things. I assume this is because the app is calling push internally somewhere during the `install` process? Halp.

src/Microcosm.js
```diff
@@ -95,6 +101,8 @@ Microcosm.prototype = {
    * @return {any} The result of the action
    */
   push(action, params, callback) {
+    if (!this.started) throw new Error('Cannot push: Did you forget to call app.start()')
```
nhunzaker: Should there be a question mark after `app.start()` ?

src/Microcosm.js
nhunzaker: Aye, it is. It would be called in the `state` plugin.

This execution order isn't intuitive. I'm going to think more on that. In the mean time, I think this is sufficient.

nhunzaker:
Looks good. One final thing: what about throwing `started:false` on the prototype?

greypants:
@nhunzaker updated.

nhunzaker:
Looks good to me!

nhunzaker:
:+1:

PR #111, opened 2016-02-26
Precompile transactions for 3-5x performance

nhunzaker:
Currently, the slowest part of the Microcosm dispatch process is the invocation of the `register` method that Stores can implement to subscribe to an action:

```javascript
import userActions from 'actions/user'

const UserStore = {
    getInitialState() {
        return []
    },
    register() {
        // Runs every time an is dispatched
        return {
            [userActions.add]: (users, record) => users.concat(record)
        }
    }
}
```

This PR adds a registry to Microcosm instances that memoizes this process. 

```shell
# Before
Time to dispatch 5000 actions: 14.224ms
Pushed 10000 actions in 84.615ms (average of 0.0085ms)
```

```shell
# After
Time to dispatch 5000 actions: 3.667ms
Pushed 10000 actions in 49.937ms (average of 0.0050ms)
```

Pretty cool! The simple SVG example is able to get up to about 37,000 transactions before it dips below 30 FPS. If you didn't need animation-grade performance you could probably get as high as 50,000.

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -1,14 +1,6 @@
-/**
- * Send
- * Communicate an action to a store. This is how Microcosm determines
- * how transaction parameters should update state.
- */
-
 import { formatTag } from './tag'
-import { isFunction } from './type-checks'
-import { get } from './update'
 
-export default function send (key, store, state, type, payload) {
+function generate (key, store, type) {
   let handler = store[type]
 
   if (handler === undefined && store.register) {
@@ -21,5 +13,11 @@ export default function send (key, store, state, type, payload) {
     handler = registrations[type]
   }
 
-  return isFunction(handler) ? handler.call(store, get(state, key), payload, state) : handler
+  return handler
+}
+
+export default function compile (entries, type) {
+  return entries.map(function ([ key, store ]) {
+    return { key, store, handler: generate(key, store, type) }
+  })
 }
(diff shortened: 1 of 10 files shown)


src/compile.js
```diff
@@ -1,14 +1,6 @@
...
- * Communicate an action to a store. This is how Microcosm determines
- * how transaction parameters should update state.
- */
-
 import { formatTag } from './tag'
-import { isFunction } from './type-checks'
-import { get } from './update'
 
-export default function send (key, store, state, type, payload) {
+function generate (key, store, type) {
```
ltk: I find this function name confusing. Doesn't seem like you're really generating anything here.

ltk:
Is `compile` the right word here? or is `memorize` more descriptive. just naming questions, otherwise real solid improvement :+1:

nhunzaker:
@ltk I can get behind `memorize`. Not sure where my head was at when I wrote `compile`

src/memorize.js
```diff
@@ -21,5 +13,11 @@ export default function send (key, store, state, type, payload) {
...
 
-  return isFunction(handler) ? handler.call(store, get(state, key), payload, state) : handler
+  return handler
+}
+
+export default function memorize (entries, type) {
+  return entries.map(function ([ key, store ]) {
+    return { key, store, handler: getHandler (key, store, type) }
+  })
 }
```
nhunzaker: @ltk How's `getHandler` and `memorize` feel?

PR #113, opened 2016-03-07
Reduce memory footprint even lower

nhunzaker:
Another fun one. Instead of storing the children of a history tree node as an array, I store it as a linked list on the existing node object. This prevents an array allocation. In order to calculate children, I've set up a getter that returns an arrayified version of this linked list.

The result is drastically reduced tree operations, and a much lighter memory footprint overall. Also, this improves performance so much that I increased the test benchmarks from 10,000 to 50,000 nodes.

## Before

```
Nodes  ::append()  ::root()  ::reduce(merge)  ::size()  ::prune()  Memory Growth
-----  ----------  --------  ---------------  --------  ---------  -------------
50000  0.0011ms    0.0250ms  11.77ms          0.03ms    3.98ms     0.812%

18.61ms to dispatch 50000 actions (50 samples)
- 162.70mbs of memory
- 13.89 minutes of history

Pushed 10000 actions in 40.222ms (average of 0.0040ms)
```

## After

```
Nodes  ::append()  ::root()  ::reduce(merge)  ::size()  ::prune()  Memory Growth
-----  ----------  --------  ---------------  --------  ---------  -------------
50000  0.0003ms    0.0220ms  7.08ms           0.03ms    1.69ms     1.086%

14.30ms to dispatch 50000 actions (50 samples)
- 58.43mbs of memory
- 13.89 minutes of history

Pushed 10000 actions in 38.643ms (average of 0.0039ms)
```

Memory growth is a _little_ higher because the linked list is a little sticker in GC. This isn't an issue because it does eventually clear, but the benchmark doesn't capture it. In either case, memory usage is 1/3 of the original.

### DIFF
--- a/benchmarks/dispatch-performance.js
+++ b/benchmarks/dispatch-performance.js
@@ -11,8 +11,8 @@ var Microcosm   = require('../dist/Microcosm')
 var Transaction = require('../dist/Transaction')
 var time        = require('microtime')
 
-var SIZE        = 10000
-var SAMPLES     = 500
+var SIZE        = 50000
+var SAMPLES     = 50
 
 var app = new Microcosm({ maxHistory: Infinity })
 
@@ -45,9 +45,11 @@ app.start()
  * `push` takes anywhere from 0.5ms to 15ms depending on the sample range.
  * This adds up to a very slow boot time!
  */
+var startMemory = process.memoryUsage().heapUsed
 for (var i = 0; i < SIZE; i++) {
   app.history.append(new Transaction(action, true))
 }
+var endMemory = process.memoryUsage().heapUsed
 
 /**
  * Warm up `::push()`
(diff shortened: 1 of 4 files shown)


benchmarks/dispatch-performance.js
```diff
@@ -11,8 +11,8 @@ var Microcosm   = require('../dist/Microcosm')
 var Transaction = require('../dist/Transaction')
 var time        = require('microtime')
 
-var SIZE        = 10000
-var SAMPLES     = 500
+var SIZE        = 50000
+var SAMPLES     = 50
```
nhunzaker: 50 seemed sufficient.

src/Tree.js
```diff
@@ -75,16 +75,30 @@ Tree.prototype = {
...
 
-    node.depth = parent.depth + 1
+    parent.next = this
   }
+}
+
+Node.prototype = {
+  depth: 0,
+
+  get children() {
```
nhunzaker: This will require IE9+ Browser support. We only use this for dev tools. What do you think?

PR #114, opened 2016-03-21
Add valueOf option to Connect

nhunzaker:
This is an experimental option for the Connect addon that enables it to
call `valueOf` on the value returned from a computed property. This
should open up a number of possibilities for writing query DSLs.

Basically, I want to support the following:

```javascript
var ViewModel = Connect(function (props) {
    return {
      users: Users.where({ job: 'Programmer' }).limit(10)
    }
})
```

In this example, we can provide perf optimizations to the query under the hood without exposing it to the user.

### DIFF
--- a/src/addons/connect/index.js
+++ b/src/addons/connect/index.js
@@ -5,13 +5,15 @@ import hoistStatics   from 'hoist-non-react-statics'
 import shallowEqual   from './shallow-equal'
 
 const defaults = {
-  pure : true
+  pure    : true,
+  valueOf : false
 }
 
 export default function connect (mapStateToProps, options) {
   options = { ...defaults, ...options }
 
   const isPure = !!options.pure
+  const useValueOf = !!options.valueOf
 
   return function wrapWithConnect(WrappedComponent) {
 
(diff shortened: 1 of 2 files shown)


greypants:
I'm having a hard time seeing how this changes makes `Users.where({ job: 'Programmer' }).limit(10)` possible... could you flesh that example out a bit?

nhunzaker:
Sure! It might be implemented like

```javascript
'use strict';

let Users = {

  rules: [],

  // Clone the Users object, appending a new rule. When
  // valueOf is called, it will return a function that uses these
  // rules as filters
  apply(rule) {
    return Object.create(this, {
      rules: { value: this.rules.concat(rule) }
    })
  },

  // Sift out all users that do not match a provided map of attributes
  where(attributes) {
    let keys    = Object.keys(attributes)
    let matches = user => keys.every(key => user[key] === attributes[key])

    return this.apply(users => users.filter(matches))
  },

  // Prevent more than `count` users from returning in the query result
  limit(count) {
    return this.apply(users => users.slice(0, count))
  },

  // Reduce through all rules to generate the result of a query
  valueOf() {
    return state => this.rules.reduce((state, transform) => transform(state), state)
  }
}

let query = Users.where({ job: 'Programmer' }).limit(2).valueOf()([
  { id: 1, job: 'Project Manager' },
  { id: 2, job: 'Programmer' },
  { id: 3, job: 'Analyst' },
  { id: 4, job: 'Programmer' },
  { id: 5, job: 'Programmer' }
])

console.log(query) // [ { id: 2, job: 'Programmer' }, { id: 4, job: 'Programmer' } ]
```

This should be safe for _everything I can think of_ because [valueOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/valueOf) is just the general API for obtaining the primitive version of an object. In this case, a query function.

greypants:
:+1:

PR #115, opened 2016-04-05
Support primitive action types

nhunzaker:
This commit adds the ability for a Microcosm to process string actions. In these cases, a simple identity function is used.

This falls perfectly in line with the use case for lifecycle actions, and so they have also been transitioned into simple string keys.

This also fits the use case for a lot of simple actions, such as using a store to maintain a global menu state:

```javascript
// UI actions
const openMenu = 'OPEN_MENU'

const UIStore = {
  open(state) {
    return { ...state, open: true }
  },
  register() {
    return { [openMenu]: this.open }
  }
}

app.push(openMenu)
```


h0tl33t:
Nice -- seems like a solid change.  :+1:

greypants:
:+1:

PR #116, opened 2016-04-05
WIP: Take advantage of parent connect recalculation

### LINKED ISSUE #112: Confirm and possibly eliminate double Connect updates
If a `Connect` instance is used within another `Connect` instance we should confirm that the React component tree only renders once. This might be happening, however test coverage would be an excellent upgrade.

nhunzaker:
When nesting connects, a change in state would trigger 3 renders:

1. The parent re-renders due to connect
2. The child re-renders due to parent
3. The child re-renders due to connect

This commit internalizes Diode as Emitter. It modifies `subscribe` to
allow for a reversed event subscription pattern and always tries to
recalculate state when a component receives properties.

So basically, if a parent wrapped in connect changes, it'll take advantage of the fact that the children wrapped on connect might re-render anyway. This prevents duplicative rendering.

Fix #112: Confirm and possibly eliminate double Connect updates


solomonhawk:
Interesting! I'm not sure I fully understand how this works.

nhunzaker:
@solomonhawk I should probably add some comments, but it might be easier to talk in person if you've got a minute. It might be helpful for me to figure out how to explain this as well.

nhunzaker:
@solomonhawk but the gist of it is that `componentDidMount` occurs in reverse. So the event subscriptions would go `child -> parent`.

nhunzaker:
This will probably land soon in 10.x

PR #117, opened 2016-04-12
Add concept of workflows

nhunzaker:
This commit adds a way to bundle together multiple actions into a common
"workflow". While a workflow is underway, it shares a common closure
with all subsequent workflow calls. This makes it easy to cancel prior
actions (such as outstanding networking requests for irrelevant
information).

Additionally, it simplifies Transaction in anticipation for a lower
level API for operating on transactions that will eventually be exposed
to the public.

### Example One: Automatically Cancelling Prior Requests

```javascript
function sendQuery(query) {
    return API.search.get({ query })
}

function getSearchResults(push, done) {
    // Do work before processing actions
    let lastRequest = null

    return function (query) {
        if (lastRequest) {
            lastRequest.cancel()
        }
        lastRequest = push(sendQuery, query, done)
    }
}

app.workflow(getSearchResults, { keyword: 'foo' })

// This next call will cancel the prior search result
app.workflow(getSearchResults, { keyword: 'bar' })
```

### Example Two: Animation

```javascript
import easing from 'easing-function'

const updateProgressBar = "update_progress_bar"

function animate (push, done) {
  let frame = null

  return (from, to) => {
    cancelAnimationFrame(frame)

    frame = requestAnimationFrame(function loop() {
      if (from > to) {
        done(null, complete)
      } else {
        push(updateProgressBar, from++)
        requestAnimationFrame(loop)
      }
    })
  }
}

app.workflow(animate, [ 0, 100 ])
```

## What needs answered

Are there pitfalls I am missing? Is there a way that we could just use `push`? This is technically non-breaking, but maybe for a `10.0.0` we could consolidate them?

Is `workflow` a good name? Would something like `pipeline` or `stream` be better?

animate


ltk:
Really cool. Love the idea.

I'm a little concerned about what this looks like from a code-reuse perspective in larger projects. I can foresee circumstances where a naive developer uses an existing workflow for a new feature without really understanding how it will impact existing features. It would be great to reuse

**RE naming:**
`pipeline` and `stream` are way worse names IMO. I think `workflow` provides a pretty good metaphor. Another option would be `bundle`, as this functionality conceptually bundles together a number of actions.

**RE ways of doing this just using push:**
I like the way this is currently implemented, but just as a thought experiment:

What if instead of relying on global tagging for workflows/bundles, you could manually instantiate them, and pass them around wherever you need to operate with them again? This could potentially just be passed as an optional last argument to `push`. Any time a bundle is passed to `push`, it doesn't actually push any action. It just invokes the returned function from the bundle with the current action as an argument.

Using straight-up `push` with this instantiated-bundle idea could be neat as it opens possibilities related to pushing different actions as part of the same bundle.

See (very) pseudo-code below

```js
function sendQuery(query) {
    return API.search.get({ query })
}

function cancelPreviousRequests(push) {
  let lastRequest = null

  return function (action, done) {
        if (lastRequest) {
            lastRequest.cancel()
        }
        lastRequest = push(action.type, action.params, done)
    }
}

// Manual instantiation of a bundle.
// Then bundling could rely on reuse of the same, portable bundle, rather than name-matching.
// Allows for the bundling logic to be reused elsewhere in isolation.
const prevRequestCancelingBundle = app.createBundle(cancelPreviousRequests)

app.push(sendQuery, { keyword: 'foo' }, prevRequestCancelingBundle)

// Within `push`, again totally pseudo

if (bundle) {
  bundle(action)
} else {
  // Actually push the action.
}

// `createBundle` would look something like
createBundle(bundler) {
  return bundler(this.push)
}
```

PR #122, opened 2016-08-02
Update linting process

ltk:
- Fixes a bug where eslint was run against no files.
- Adds linting of JS code in markdown documents.
- Fixes an unused var issue in a test.
- Ignores another unused var issue in a different test. (I can't find a
great way to ignore unused vars assigned to `_`.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "src/microcosm.js",
   "scripts": {
     "build": "make all",
-    "lint": "eslint {src,test,examples}/**/*.{js,jsx}",
+    "lint": "eslint {src,test,examples}/**",
     "test": "ava",
     "test:cov": "nyc --cache --reporter=html ava",
     "release": "make release"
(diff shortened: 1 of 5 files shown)


ltk:
I also just squeed some badge updates to the README in here as well.

package.json
```diff
@@ -6,7 +6,7 @@
   "main": "src/microcosm.js",
   "scripts": {
     "build": "make all",
-    "lint": "eslint {src,test,examples}/**/*.{js,jsx}",
+    "lint": "eslint {src,test,examples}/**",
```
nhunzaker: Whhat. Seriously? But I was getting linting errors before.... Good catch either way.

Jeez.

src/getStoreHandlers.js
```diff
@@ -1,5 +1,7 @@
 function format (string) {
+  /*eslint-disable no-unused-vars*/
   const [ _, action, state ] = `${ string }`.match(/(\w*)\_\d+\_(\w*)/, ' ') || []
+  /*eslint-enable no-unused-vars*/
```
nhunzaker: This rule is a bummer. So often we only use a second argument, particularly for general event publishers. I wish it could make an exception for vars/arguments that start with `_var`, like in Erlang.

package.json
ltk: very odd. eslint version difference perhaps? guess that would only happen if you were somehow using a global copy.

nhunzaker:
👍 Neato

src/getStoreHandlers.js
ltk: Yeah, there's some discussion around this here: https://github.com/eslint/eslint/issues/1939

Could probably write a custom rule for this.

PR #129, opened 2016-08-11
Add AJAX recipe to documentation

nhunzaker:
Easy enough. cc @ltk 

I haven't covered cancellation or progress. These are special cases that I want to give their own recipes. Thoughts?


ltk:
noice :+1:

PR #130, opened 2016-08-11
Presenter docs

nhunzaker:
Adds docs for presenters.

This needs copy editing. I also need feedback on the API for Form and Presenter.

### DIFF
--- a/examples/chatbot/app/actions/messages.js
+++ b/examples/chatbot/app/actions/messages.js
@@ -1,4 +1,4 @@
-export function send (message) {
+export function send ({ message }) {
   var request = new XMLHttpRequest()
 
   return function (action) {


docs/api/form.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+Form inputs are serialized to JSON upon submission using
+[`form-serialize`](https://github.com/defunctzombie/form-serialize).
+
+## Props
+
+### intent
+
+A string value to broadcast out upon submission. This should
```
ltk: What does "broadcast out" mean? It wasn't initially clear to me how this was working when I saw the form component.

nhunzaker: Thank you. This is a really bad term. Basically this is going to call `Presenter.send` on the associated presenter, which will bubble up through the Presenter hierarchy and see if any of them implement the associated property.

PR #131, opened 2016-08-15
Add setup method

nhunzaker:
Tapping into `super` within the constructor when extending Microcosm feels too low level. What do you think about adding a setup step? Personally, I think it indicates the intent much clearer.

Replaces:

```javascript
class SolarSystem extends Microcosm {
  constructor(options) {
    super(options)
    this.addStore('planets', Planets)
  }
}
```

With:

```javascript
class SolarSystem extends Microcosm {
  setup() {
    this.addStore('planets', Planets)
  }
}
```

This also eliminates a friction point: remembering to pass `options` into `super()`.

### DIFF
--- a/src/microcosm.js
+++ b/src/microcosm.js
@@ -40,7 +40,7 @@ export default class Microcosm extends Emitter {
     // Standard store reduction behaviors
     this.addStore(MetaStore)
 
-    // Microcosm is now ready. Call the initialize method
+    // Microcosm is now ready. Call the setup lifecycle method
     this.setup()
   }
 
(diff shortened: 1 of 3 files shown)


mackermedia:
I like it. Explicit++ 👍

src/microcosm.js
```diff
@@ -39,6 +39,18 @@ export default class Microcosm extends Emitter {
 
     // Standard store reduction behaviors
     this.addStore(MetaStore)
+
+    // Microcosm is now ready. Call the initialize method
```
greypants: > Call the ~~initialize~~ setup method.

Or rename setup to `initialize`?

greypants:
👍 I like the change.

src/microcosm.js
nhunzaker: Oops. Duh. I called this initialize before, but then changed it to `setup` because JS class libraries have a history of using `init` and `initialize` sort of like a constructor.

PR #137, opened 2016-09-13
Warn when returning repo.state, allow viewmodel function

nhunzaker:
Sometimes a presenter just wants to subscribe to all state changes. We can't just return `this.repo.state`, because the view model is only recalculated when a presenter receives new props.

So here's my solution: support returning a function that is given the current repo state:

```javascript
class MyPresenter extends Presenter {
  viewModel() {
    return state => state
  }
  render() {
    return <p>Test</p>
  }
}
```

Or, to quickly extract several keys:

```javascript
class MyPresenter extends Presenter {
  viewModel() {
    return ({ dogs, cats }) => ({ dogs, cats })
  }
  render() {
    return <p>Test</p>
  }
}
```

We'll also now warn if `this.repo.state` is returned from the view model. What do you think?

### DIFF
--- a/docs/api/presenter.md
+++ b/docs/api/presenter.md
@@ -60,6 +60,24 @@ planets from the Microcosm instance provided to it via the `repo`
 prop. This is available as state, which the Presenter can send into a
 child component.
 
+### Listening to all state changes
+
+While we do not recommend it for large projects, some times it's simply easier
+to subscribe to all repo changes. `viewModel` can also return a function, which
+will be called with state:
+
+```javascript
+class PlanetsPresenter extends Presenter {
+  viewModel(props) {
+    return state => state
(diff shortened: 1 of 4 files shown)


nhunzaker:
I have mixed thoughts on returning a function. On one hand, it's very ergonomic. On the other, it totally invites unexpected rendering inefficiencies. Still I also can't help but wonder if there are other neat things you could do here, like make a view model module:

cwmanning:
👍

cwmanning:
@nhunzaker easy to add those alternate code examples from the PR description into the docs?

nhunzaker:
Yess. I sure can.

nhunzaker:
@cwmanning done.

PR #147, opened 2016-09-22
withIntent addon

nhunzaker:
We have a Form add-on that connects to Presenters. @cwmanning has been utilizing a `withIntent` higher-order component that is better suited for buttons, or other components:

```javascript
const DestroyButton = withIntent(function ({ send }) {
  return (
    <button onClick={() => send('destroy-thing')}>
      Nuke this thing
    </button>
  )
})

class Users extends Presenter {
  register() {
    return {
      'destroy-thing': repo => repo.push(destroyThing)
    }
  }

  render() {
    // Kind of silly, probably better to use a callback prop here, but imagine
    // this is deep in the component tree
    return <DestroyButton />
  }
}
```

## One Question

Should we be more restrictive with passing down send? Something like `withIntent("intent-name", Component)`. We could add extra validations to make sure the intents are connected properly, and it might avoid some goofy dynamic code:

```javascript
const DestroyButton = withIntent('destroy-thing', function ({ send }) {
  return (
    <button type="button" onClick={send}>
      Nuke this thing
    </button>
  )
})
```

I'm kind of partial to this, but I could see cases where you would want to switch intents. Still, I wonder for those cases how much of the associated component would also change based on that switch (like a menu toggle changing from open to close). In those cases, I'd rather just have two buttons.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -25,3 +25,4 @@ start. Beyond that, check out [the example apps](../examples).
 
 1. [Presenter](api/presenter.md)
 2. [Form](api/form.md)
+3. [withIntent](api/with-intent.md)
--- a/docs/api/with-intent.md
+++ b/docs/api/with-intent.md
@@ -1,0 +1,47 @@
+# withIntent(Component)
+
+1. [Overview](#overview)
+2. [API](#api)
+
+## Overview
+
+A higher order component that can be used to connect a component deep
+within the component tree to its associated presenter. This is useful
+for requesting work from Presenters where passing down callbacks as
+props may otherwise be exhaustive.
+
+```javascript
+import React from 'react'
+import withIntent from 'microcosm/addons/with-intent'
+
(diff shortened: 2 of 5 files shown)


cwmanning:
👍 looks great! 

I'm not sure I see a need to be restrictive, at least initially. The validations would be nice, but it seems like it would make it a little more complicated to implement.

> I could see cases where you would want to switch intents

Could most of these cases be solved by sending an argument(s)? A menu button might `this.props.send('toggle', isOpen)` but interested if I'm missing something here.

nhunzaker:
@cwmanning Sounds good to me!

Actually, the way I think I would use this would be to make a general `IntentButton` component that operated something like:

```javascript
<IntentButton intent="toggle" value={!open}>{open ? 'Close' : 'Open'}</IntentButton>
```

To that end, `withIntent` is already tucked away.

Issue #148, opened 2016-09-22
Presenter::present


nhunzaker:
What if we gave Presenter a special render method that passed in the current view model? Maybe we could even have `viewModel` assign to a `model` state namespace or something to avoid collisions with `this.setState`:

```javascript
class Users extends Presenter {
    setup(repo) {
        repo.push(getUsers)
    }

    viewModel() {
        return { users: state => state.users }
    }

    present(props, { users }) {
        return <UsersTable users={users} />
    }
}
```

What do you think?

nhunzaker:
Another alternative is to hijack `render`, though I'm not super extend about that. `present` would essentially just do:

```javascript
class Presenter { 
    // ....
    present () {
        return this.render(this.props, this.state)
    }
}
```

PR #149, opened 2016-09-23
Add view method and alias for viewModel

nhunzaker:
I think, hopefully, this is the final touch on `10.0.0`. This PR adds a special rendering method for Presenters: `view`. When implemented, it accepts the current view model:

```javascript
class Greeter extends Presenter {

  model ({ greeting })
    return {
      message: state => greeting + ", " + state.location
    }
  }

  view ({ message }) {
    return <p>{message}</p>
  }

}

const repo = new Microcosm()

repo.replace({ location: 'World' })

DOM.render(<Greeter repo={repo} greeting="Hello" />, el)
// "Hello, World"
```

I've also added an alias for `viewModel`: `model`.

I believe there is an improvement to ergonomics here, and that this eliminates some of the strangeness with accessing `this.state` to get at a view model in `render`.

Users will always still have the option to just use `render`.

---

#148

### DIFF
--- a/examples/painter/app/views/row.jsx
+++ b/examples/painter/app/views/row.jsx
@@ -1,18 +1,19 @@
 import React from 'react'
 import withIntent from '../../../../src/addons/with-intent'
 
-withIntent(function Cell ({ x, y, active, send }) {
+const Cell = function ({ x, y, active, onClick }) {
   const color = active ? 'black' : 'white'
-  const paint = () => send('paint', { x, y })
 
-  return <rect key={x} x={x} y={y} onClick={paint} fill={color} width="1" height="1"/>
-})
+  return (
+    <rect x={x} y={y} onClick={onClick} fill={color} width="1" height="1"/>
+  )
+}
 
-export default function Row ({ cells, y, onClick }) {
+export default withIntent(function Row ({ cells, y, send }) {
 
   return (
     <g key={ y }>
-      { cells.map((active, x) => Cell({ x, y, active, onClick })) }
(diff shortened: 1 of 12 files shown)


examples/painter/app/views/row.jsx
```diff
@@ -1,16 +1,19 @@
 import React from 'react'
+import withIntent from '../../../../src/addons/with-intent'
 
-function Cell ({ x, y, active, onClick }) {
+const Cell = function ({ x, y, active, onClick }) {
```
greypants: why the switch to assignment?

src/addons/presenter.js
```diff
@@ -210,10 +220,14 @@ export default class Presenter extends Component {
     console.warn(`No presenter implements intent “${ intent }”.`)
   }
 
-  render () {
+  view (model) {
     return this.props.children ? React.Children.only(this.props.children) : null
```
greypants: what's this helper do?  `React.Children.only`

examples/painter/app/views/row.jsx
nhunzaker: Ack. Bad flow (and a good catch). I wrapped `Cell` in `withIntent`, but then found it cleaner just to do it in `Row`. I'll revert this.

Thank you!

src/addons/presenter.js
nhunzaker: It makes sure there's only one child. Otherwise it throws with a useful error.

https://facebook.github.io/react/docs/top-level-api.html#react.children.only

Kind of makes me wonder if it's necessary...

nhunzaker: *Or if React will just do the same thing.

PR #157, opened 2016-10-14
Do not support string key access on domain dispatch

nhunzaker:
Basically don't let this happen:

```javascript
const repo = new Microcosm()

repo.addDomain('test', {
  test() {
      console.log("bad bad bad!")
  }
})

repo.push('test')
```

This is important because I want to support broadcasting actions as intents within `withIntent` and presenters.


greypants:
👍

PR #168, opened 2016-10-25
Subscribe to repo on mount

nhunzaker:
https://twitter.com/dan_abramov/status/790581793397305345:

> Don’t subscribe to data sources in constructor or componentWillMount(). Use componentDidMount(), teardown in componentWillUnmount().

So we don't, anymore. I've also removed `this.pure` in favor of `this.isImpure()`.

### DIFF
--- a/src/addons/presenter.js
+++ b/src/addons/presenter.js
@@ -4,8 +4,6 @@ import merge from '../merge'
 import tag from '../tag'
 import shallowEqual from '../shallow-equal'
 
-const EMPTY = {}
-
 function wrappedRender () {
   return (
     <PresenterContext {...this.props} presenter={this}>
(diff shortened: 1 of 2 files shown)


src/addons/presenter.js
```diff
@@ -21,8 +19,8 @@ function wrappedRender () {
  */
 class Presenter extends React.Component {
 
-  constructor() {
-    super(...arguments)
+  constructor(props, context) {
+    super(props, context)
```
greypants: 👍 for clarity

greypants:
That's a fun twitter thread.

ltk:
:+1:

nhunzaker:
For whatever it's worth, each Presenter gets a fork of the parent repo. Subscriptions are localized to that repo. I'm not actually sure we _can_ leak event subscriptions.

nhunzaker:
Thanks everyone!

PR #169, opened 2016-10-25
Break <Form> submit out of onSubmit

greypants:
There are times where you might want to submit a form programmatically from an external component, or a special event, like on shift+enter (not enter) in message field. `event.preventDefault()` isn't needed in this case, so that's left in `onSubmit`, leaving `submit` callable without it.

Example:

```js
    <div>
      <Form ref={form => this.form = form} intent="updateMessage">
      </Form>
      <button type="button" onClick={() => this.form.submit()}>Save</button>
    <div>
```

Didn't write tests yet.


src/addons/form.js
```diff
@@ -43,7 +43,10 @@ const Form = React.createClass({
 
   onSubmit(event) {
     event.preventDefault()
+    this.submit(event)
```
greypants: hmm... this would need `event.persist` if the event was a thing, wouldn't it 😕

src/addons/form.js
```diff
@@ -43,7 +43,10 @@ const Form = React.createClass({
...
+  },
 
+  submit() {
     const form   = event.target
```
greypants: This would need to use a ref instead of the event target.

PR #171, opened 2016-10-25
Directly submit a Form component

### LINKED ISSUE #170: Programmatically submit a Microcosm <Form>
There are times where you might want to submit a form programmatically from an external component, or a special event, like submitting on enter `keyDown` from a `<textarea />`, which isn't the default behavior. `onSubmit` calls `event.preventDefault()` isn't needed in this case, so we would need to break out a separate `submit` method, which doesn't rely on the presence of `event`.

Example:

```js
    <div>
      <Form ref={form => this.form = form} intent="updateMessage">
      </Form>
      <button type="button" onClick={() => this.form.submit()}>Save</button>
    <div>
```

I tried using refs and doing `findDOMNode(this.refs.form).submit()`, but the form submits (and refreshes the page) without triggering the internal `onSubmit` callback, which prevents default 😕

greypants:
This allows you to programmatically submit a Microcosm `<Form />`, usually via refs like so:

```js
this.refs.myCoolMicrocosmForm.submit()
```

Resolves #170


CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ npm test
 For test coverage:
 
 ```bash
-npm run test:cov:html
+npm run test:cov
```
nhunzaker: :+1:

tommymarshall:
i like yo styleeeeee

cwmanning:
Nice. Make sure to add a quick section to the docs!

PR #176, opened 2016-11-01
Make send directly available to Presenter

greypants:
This allows the method to be accessed from the `Presenter` directly.

Basically so we can do pass the `send` method from a presenter to other classes or functions in the setup method: 

```js
class MySocketyPresenter extends Presenter {
  setup(repo) {
     this.sockets = {
        chat: new IntentSocket(`graph/${id}/chat`, this.send), 
        graph: new IntentSocket(`graph/${id}/graph`, this.send)
      }
  }
}
```

### DIFF
--- a/src/addons/presenter.js
+++ b/src/addons/presenter.js
@@ -19,54 +19,6 @@ function wrappedRender () {
  */
 class Presenter extends React.Component {
 
-  static contextTypes = {
-    send : React.PropTypes.func
-  }
-
-  static childContextTypes = {
-    send : React.PropTypes.func
-  }
-
-  getChildContext () {
-    return {
-      send : this.send.bind(this)
-    }
-  }
-
-  /**
-   * Provides a way for views to send messages back to the presenter
-   * in a way that does not require passing callbacks down through the
-   * view. This method is exposed by the Presenter via the React context
-   * API.
-   *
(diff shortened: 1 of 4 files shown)


src/addons/presenter.js
```diff
@@ -19,6 +19,54 @@ function wrappedRender () {
  */
 class Presenter extends React.Component {
 
+  static contextTypes = {
+    send : React.PropTypes.func
+  }
+
+  static childContextTypes = {
+    send : React.PropTypes.func
+  }
```
nhunzaker: I had to move these because IE10 doesn't forward along statics to inherited classes :(

greypants: should I duplicate them then?

nhunzaker:
@greypants we need to keep context inside of the wrapper to support IE10, otherwise context is not transferred when you `extend` Presenter. Extremely frustrating.

I've made a change that installs `send` when the wrapper is instantiated. What do you think?

src/addons/presenter.js
```diff
@@ -33,6 +33,10 @@ class Presenter extends React.Component {
     this.setup(repo, this.props)
   }
 
+  _connectSend (send) {
+    this.send = send
```
greypants: 👍

greypants:
👍

PR #178, opened 2016-11-07
DO NOT MERGE: First pass at experimental databank addon

nhunzaker:
This is an experiment to improve the experience updating data. It implements a custom domain that allows a user to return the following pattern:

```
[ operation, key, value ]
```

For example:

```javascript
import Databank from '../../../../src/addons/databank'
import { addList, removeList } from '../actions/lists'

class Lists extends Databank {

  add (lists, params) {
    return ['put', params.id, params]
  }

  remove (lists, unwanted) {
    return ['destroy', unwanted]
  }

  register() {
    return {
      [addList]    : Lists.add,
      [removeList] : Lists.remove
    }
  }

}
```

### Outstanding thoughts

This isn't perfect. I'm already finding cases in real-world projects that need to be accounted for. But the underlying goal is to move away from doing data manipulation yourself, allowing Microcosm to handle that in a guaranteed safe way.


greypants:
First through is that it feels foreign, and somewhat opaque. I guess the main benefit is not having to pass in lists?

Personally, i'd rather have data-manipulation helper methods that ship with the library... something like this:

```
import { add, remove } from 'microcosm/addons/data-manipulation'
import { addList, removeList } from '../actions/lists'

const Lists = {
  add (lists, params) {
    return add(lists, params)
  }

  remove (lists, id) {
    return remove(lists, id)
  }

  removeAttribute(lists, prop) {
    return remove(lists, prop, { path: ['attributes']})
  }

  register() {
    return {
      [addList]    : Lists.add,
      [removeList] : Lists.remove
    }
  }
}
```

greypants:
I am on-board with your goal though

> The underlying goal is to move away from doing data manipulation yourself, allowing Microcosm to handle that in a guaranteed safe way.

❤️

nhunzaker:
@greypants word. That would be a good first step anyway. It'll be extremely hard to find the right pseodo-dsl for a Domain. Having good data management helpers is an important start.

So let's figure those out. On our projects, what are the common operations that we perform?

greypants:
`add`, `update`, `remove`, `replace` is what I've been using.

nhunzaker:
@greypants Maybe we could lean on https://github.com/kolodny/immutability-helper

greypants:
Huh. Didn't know that was a thing. I'll look into that.

nhunzaker:
@greypants I'll close this out. Let's continue investigating things like `sprout-data` and `immutability-helpers`

PR #179, opened 2016-11-11
Effects

nhunzaker:
This PR adds the concept of effects to Microcosm, a way to apply side-effects in a safe way. Whenever an action moves into a new state, it asks the history object to reconcile. After reconciliation, I've added an extra step for handling side effects. This should support a few use cases:

1. Side-effects in presenters that are the result of actions happening
2. Secondary data fetching

```javascript
const repo = new Microcosm()

const Notifications = {
  notify (repo, params) {
    alert(params.name + ' was updated!')
  }

  register() {
    return {
      [createUser] : this.notify
    }
  }
}

repo.addEffect(Notifications)

repo.push(createUser, { name: 'Billy' })
```

These can be added to Presenters too, and localized to a particular fork. So side-effects can be managed more strategically than a simple register method:

```javascript
class MyPresenter extends Presenter {
    setup (repo) {
        repo.addEffect(Effect, { options })
    }
}
```

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,6 +22,7 @@ start. Beyond that, check out [the example apps](../examples).
 1. [Microcosm](api/microcosm.md)
 2. [Domains](api/domains.md)
 3. [Actions](api/actions.md)
+4. [Effects](api/effects.md)
 
 ## Addons
 
--- a/docs/api/effects.md
+++ b/docs/api/effects.md
@@ -1,0 +1,79 @@
+# Effects
+
+1. [Overview](#overview)
+2. [Subscribing to different action states](#subscribing-to-different-action-states)
+3. [API](#api)
+
+## Overview
+
+Effects are one-time handlers that are invoked once, after the action
+moves into a different state. They are very similar to Domains,
+however their purposes is to handle side-effects. Domains are not the
+appropriate place to handle this sort of behavior, as they may
+dispatch the same handler several times during the reconciliation of
+asynchronous actions.
+
+```javascript
+// If an effect implements a setup method, it will receive options as
+// the second argument
+repo.addEffect(Effect, options)
+```
+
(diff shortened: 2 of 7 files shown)


src/getHandlers.js
```diff
@@ -0,0 +1,42 @@
+import lifecycle from './lifecycle'
+
+function format (string) {
+  /*eslint-disable no-unused-vars*/
+  const [ _, action, state = 'done' ] = `${ string }`.match(/(\w*)\_\d+[\_\w]*/) || []
+  /*eslint-enable no-unused-vars*/
```
greypants: oh, nice. Didn't realize you could do that inline. Sad that we can't make a global exception for this pattern 😞

nhunzaker: Ack! This file should get removed. Good catch!

greypants:
I like it. I need to sit with the term `Effect` for a bit, but I think this is super helpful. Minor concerns/things to think about: Domains, Presenters, and Effects are all feel super similar in that they all have `register` methods and callbacks. A newbie might wonder why they can't just use domains for this purpose, since they look the same and work similarly. Maybe just an issue of documentation.

greypants:
This probably screws up a bunch of naming internals, but I almost think this would be nice:

```js
class MyPresenter extends Presenter {
    setup (repo) {
        repo.register(Effects, { options })
    }
}
```

src/getHandlers.js
nhunzaker: Beyond that though, I want to say there's a thing you can do with `no-unused-vars` that ignores variables that start with `_`.

nhunzaker:
@greypants I hear you on the similarity. There's something there that I haven't figured out yet. I'm okay with it being a _little_ confusing for the sake of consistency, but I agree that the documentation needs to be clearer.

PR #180, opened 2016-11-12
Form transform

nhunzaker:
Just bumped into this on our hackathon project. This PR makes it much easier to format data before submitting it:

```javascript

class SchedulesNew extends React.Component {

  transform (params) {
    params.start = new Date(params.start).toISOString()
    params.end = new Date(params.start).toISOString()

    return params
  }

  render () {
    return (
        <Form intent={actions.create} onError={this.onFailure} transform={this.transform}>
          <p><label>Name <input name="name" /></label></p>
          <p><label>Start <input name="start" type="datetime-local" /></label></p>
          <p><label>End <input name="end" type="datetime-local" /></label></p>

          <footer>
            <input type="submit" />
          </footer>
        </Form>
    )
  }

}

export default SchedulesNew
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 10.7.0
 
-- Added `transform` method for processing form parameters after they
+- Added `prepare` method for processing form parameters after they
   are serialized.
 
 ## 10.6.1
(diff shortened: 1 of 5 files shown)


h0tl33t:
👍

PR #183, opened 2016-11-18
Patching on child repos should not effect parents

nhunzaker:
Forks inherit state. If a parent repo's state contains a key, it will
pass the associated value down to a child. This allows state to flow
downward through the repo network.

Unfortunately, this causes an unexpected behavior with `patch` where
consecutive actions always operate on the original patched value, and
not the new state produced by a child.


nhunzaker:
cc @greypants

Issue #184, opened 2016-11-29
Sub-domains


nhunzaker:
I think it's time that we have an actual class to describe Domains. We could use this to manage Sub-domains, which could operate relative to the key a Domain is mounted to:

```javascript
class Network extends Domain {

    setup (repo) {
        this.addDomain('nodes', Nodes) // mounted to ['network', 'nodes']
        this.addDomain('edges', Edges) // mounted to ['network','edges']
    }    

}

repo.addDomain('network', Network)
```

### Outstanding questions:

1. Do parent domains handle actions first, or after all subdomains have completed?

nhunzaker:
I think I've got this wrong. Let's keep domains simple. Right now they don't implement any introspective methods. 

Instead, let's make it so that a Microcosm can mount other Microcosms at a given key path.

greypants:
iiiiiiiiiiiiiiiiiiiiiiiiiiiinteresting. Sorta like the initial idea. Would definitely help me break up my giant `graph` domain, which manages nodes, edges, loading state, ui, and some other stuff.

My gut says that children should finish their work first, then parents.

nhunzaker:
Closing this out again. Sub-domains work well when the state tree is highly nested, but I think domains probably need to stay flat. Nesting largely is necessary for convenience of access, which we can fix by adding a querying API.

PR #185, opened 2016-11-29
Add concept of sub-domains

nhunzaker:
I want to slowly move towards Domains just being Microcosms mounted at paths to other Microcosms:

```javascript
import Domain from 'microcosm/domain'

class Node {
  getInitialState() { return [] }
}

class Edge {
  getInitialState() { return [] }
}

class Network extends Domain {
  setup () {
    this.addDomain('nodes', Nodes)
    this.addDomain('edges', Edges)
  }
}

const repo = new Microcosm()

repo.addDomain('network', Network)

expect(repo.state.network.edges).toEqual([])
expect(repo.state.network.nodes).toEqual([])
```

I also updated error messages to correctly encounter non-function keys (and respond with null):

```
Unable to register `action.done` at domain path `["network", "nodes"]`. Handlers must be
functions, instead got `undefined`. Check the register method for this domain.
```

### DIFF
--- a/docs/api/domains.md
+++ b/docs/api/domains.md
@@ -192,18 +192,18 @@ sub-domains for themselves.
 ```javascript
 import Domain from 'microcosm/domain'
 
-class Node {
+class Nodes {
   getInitialState() { return [] }
 }
 
-class Edge {
+class Edges {
   getInitialState() { return [] }
 }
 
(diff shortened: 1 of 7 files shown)


nhunzaker:
Closing this out after I mused over it a while. See #184

Issue #172, opened 2016-10-26
Throw a warning if you try to use `render` in a presenter?


greypants:
`Presenter` requires a `view` method instead of `render`. We should warn users upgrading.

nhunzaker:
Maybe not a deprecation, but def a warning.

PR #193, opened 2016-12-05
Add warning in render method for Presenter

nhunzaker:
Name says it all!

Fixes https://github.com/vigetlabs/microcosm/issues/172

---

cc @greypants

### DIFF
--- a/src/addons/presenter.js
+++ b/src/addons/presenter.js
@@ -26,6 +26,10 @@ class Presenter extends React.Component {
 
     // Allow overriding render, generate the context wrapper
     // upon instantiation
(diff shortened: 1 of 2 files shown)


src/addons/presenter.js
```diff
@@ -26,6 +26,10 @@ class Presenter extends React.Component {
 
     // Allow overriding render, generate the context wrapper
     // upon instantiation
+    if (this.render !== Presenter.prototype.render) {
+      console.error('Presenter::render is a protected method. Instead of overriding it, please use Presenter::view.')
```
greypants: nice

PR #197, opened 2017-01-05
Allow custom repo behavior in Presenter

nhunzaker:
This commit adds an extendable getRepo method to the Presenter
prototype that allows custom repo assignment behavior. For example, if
we wanted to avoid the default forking behavior:

```javascript
class NoFork extends Presenter {
  getRepo (repo) {
    return repo
  }
}
```

### DIFF
--- a/docs/api/presenter.md
+++ b/docs/api/presenter.md
@@ -276,3 +276,13 @@ class HelloWorldPresenter extends Presenter {
   }
 }
 ```
+
+### getRepo(repo, props)
+
+Runs before assigning a repo to a Presenter. This method is given the
+parent repo, either passed in via `props` or `context`. By default, it
+returns a fork of that repo, or a new Microcosm if no repo is
+provided.
+
+This provides an opportunity to customize the repo behavior for a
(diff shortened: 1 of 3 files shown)


docs/api/presenter.md
```diff
@@ -276,3 +276,22 @@ class HelloWorldPresenter extends Presenter {
...
+provided.
+
+This provides an opportunity to customize the repo behavior for a
+particular Presenter. For example, to circumvent the default Presenter
+forking behavior:
+
+```javascript
+class NoFork extends Presenter {
+  getRepo (repo) {
+    return repo
```
cwmanning: what about an inline comment clarifying that this is a new Microcosm?

cwmanning:
👍

docs/api/presenter.md
nhunzaker: This is the same repo given to the Presenter (unless no repo is given). Is that what you mean?

nhunzaker: Cool

PR #199, opened 2017-01-05
Payload definition

### LINKED ISSUE #198: Pushing action with no params, calls handlers with null
Pushing action with no params, calls handlers with `null` as the second param (after `state`), instead of `undefined`. I noticed the same behavior using `send` from a `withIntent` component.

Code demonstrating the issue:

```js
const Microcosm = require("microcosm").default
const repo = new Microcosm()

const domain = {
  getIntialState() {
    return {
      someToggle: false
    }
  },
    
  register() {
    return {
      'toggleSomeToggle': (state, someToggle) => {
        console.log('someToggle === ' + someToggle)
        return {...state, someToggle}
      }
    }
  }
}

repo.addDomain('domain', domain)

repo.push('toggleSomeToggle')
```

Yields:

```js
"someToggle === null"
```

Demo here: https://runkit.com/greypants/microcosm-issue-198

Explicitly sending along `undefined` yields the same result

```js 
repo.push('toggleSomeToggle', undefined) // toggleSomeToggle === null
```

This prevents setting defaults param defaults like so:

```js
  'toggleSomeToggle': (state, someToggle = true) => { ... }
```

Because `someToggle` is always defined as `null` if no params are passed.

nhunzaker:
Action payloads may now intentionally be set to undefined. We'll also make them undefined by default.

---

Should fix #198

### DIFF
--- a/src/action.js
+++ b/src/action.js
@@ -61,7 +61,7 @@ inherit(Action, Emitter, {
     this.type = this.behavior[type]
     this.disposable = DISPOSABLE.hasOwnProperty(type)
 
-    if (payload !== undefined) {
+    if (arguments.length > 1) {
       this.payload = payload
     }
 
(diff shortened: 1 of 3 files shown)


src/action.js
```diff
@@ -16,7 +23,7 @@ export default function Action (behavior) {
 
 inherit(Action, Emitter, {
   type      : null,
-  payload   : null,
+  payload   : undefined,
```
greypants: 🎉

PR #200, opened 2017-01-06
Updates commit documentation to explain 'public' vs. 'private' concept

mackermedia:
Thoughts?


nhunzaker:
👍 . This is much clearer. We should go through the docs some time and see if they'd benefit from similar updates.

nhunzaker:
Updated site too:
http://code.viget.com/microcosm/api/domains.html#commitnext-staged

PR #204, opened 2017-01-10
Pass Presenter send method along with model

### LINKED ISSUE #203: Pass `send` down to direct children of Presenter
That way we don't have to wrap everything in withIntent

nhunzaker:
We found ourselves wrapping the direct children of the Presenter component with `withIntent` quite often. This PR configures the Presenter to pass down `send` along with the model.

Fixes #203

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 11.2.0
+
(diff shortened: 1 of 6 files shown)


h0tl33t:
👍

nhunzaker:
That was fast

mackermedia:
👍

PR #206, opened 2017-01-10
withIntent should properly set displayName

### LINKED ISSUE #205: withIntent should set up displayName

nhunzaker:
That way it is easily selectable within Enzyme. Like:

```javascript
let name = enzymeWrapper.find('withIntent(ComponentName'))
```

Fixes #205


mackermedia:
:+1:

PR #220, opened 2017-02-15
Updates to Effects documentation.

mackermedia:
* updates `patchQuery` & `addPlanet` action var names to match import and usage
* adds file location description as comment (@nhunzaker is there a better path to display?)
* demonstrates adding an Effect to the repo/microcosm


nhunzaker:
👍 Thanks!

PR #221, opened 2017-02-16
Reset and patch reject when deserialize fails. Add new test helper

nhunzaker:
Reset and patch return an action. This PR sets those actions up so that they are rejected whenever deserializing their payloads fail.

Deserialize still raises an exception. 

I've also added a new test helper for quickly testing repo state in a safe way.

### DIFF
--- a/src/addons/jest-matchers.js
+++ b/src/addons/jest-matchers.js
@@ -1,4 +1,4 @@
-import { tag, getRegistration } from '../microcosm'
+import Microcosm, { Action, tag, get, getRegistration } from '../microcosm'
 
 expect.extend({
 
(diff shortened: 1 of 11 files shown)


src/addons/jest-matchers.js
```diff
@@ -18,6 +18,11 @@ expect.extend({
   },
 
   toHaveStatus (action, status) {
+    if (action instanceof Action === false) {
+      throw new TypeError("toHaveStatus expects an Action. Received " +
+                          (action ? "a " + action.constructor.name : action) + ".")
```
mackermedia: You want to return null if no action on the second ternary return value? Might be easier to read by explicitly having `null` there?

nhunzaker: I want to account for other values, like `undefined`, a string, or a number, or something else. This at least calls `toString` on it.

mackermedia:
👍

src/addons/jest-matchers.js
mackermedia: ah okay 👍

nhunzaker: Ah.. hmm. I see your point though. It'll be confusing to read:

```
expect("foobar").toHaveStatus("closed")
toHaveStatus expects an Action. Received "foobar".
```

Instead of:

```
"toHaveStatus expects an Action. Received a String."
```

What's better in your opinion?

PR #223, opened 2017-02-20
Fixes 404'd page for effect.smd => .html

peterramsing:
I'm not sure if this is where to fix this–but figured I'd see if this helps.

Cheers


nhunzaker:
Ah! Good catch! I'll ship this over to get an immediate fix in. These pages get generated from markdown files in the docs folder. We need to replace cases where they link to other markdown files with their html counterparts.

Thanks for the fix!

peterramsing:
@nhunzaker I glanced at the docs directory and maybe I didn't look close enough–if you show me where the source is I can make the updates there, too. I can also update the "CONTRIBUTING.md" file to have some notes for updating docs if you'd like.

Thanks for being epic!

nhunzaker:
@peterramsing Awesome! The docs are here:

https://github.com/vigetlabs/microcosm/tree/master/docs

We take them and convert them into HTML using Pandoc, as a part of a Makefile:

https://github.com/vigetlabs/microcosm/blob/master/site/Makefile

So the trick here is that we need to add a pass that replaces `*.md` with `*.html` in our links. 

This is also a good opportunity to make sure our contributor documentation is up to date, and see if we can make this process easier.

PR #227, opened 2017-02-22
Query Addon

nhunzaker:
This PR adds a new add-on: Query. Query is a data querying library designed specifically for Microcosm. In future releases, it will be hook more tightly into Microcosm to deliver highly efficient data queries. It is designed for use with the Presenter addon, though I imagine it would be useful in other places:

```javascript
model () {
  var q = new Query()

  return {
    blocks  : q.get('blocks').filter(block => block.parent == null),
    theme   : q.get('styles').tap(applyTheme),
    step    : q.get('ui.step'),
    message : q.get('message')
  }
}
```

It's also quite small, for the time being. Current query features include:

get - Pluck out a key path from state
map - Map over a list of data
filter - Filter a list of data
join - Filter down to records matching the value in another key path
tap - Arbitrarily call a function on the result
at - Extract a specific index from a list of data

Each query is memoized and will always return the same result if
shallowly equal to the last.


test/query.test.js
```diff
@@ -0,0 +1,229 @@
...
+  it('returns undefined if out of the upper boundary', function () {
+    let query = new Query().get('planets').at(10)
+
+    expect(query.compute(state)).toBeUndefined()
+  })
+})
+
+describe('join', function () {
+  it('filters by a criteria within the dataset', function () {
+    let query = new Query().get('planets').join('id', 'meta.selected').at(0).get('id')
```
greypants: faaancy

greypants: but also, that line is sort of horrible to read and understand, and should probably never be written in real life...

nhunzaker: Yeah... we need to establish some additional query functions to more quickly get at what we want. It's definitely not there yet.

greypants: `join` is a confusing name to me, especially since `Array.prototype.join` is a thing, and I'd sort of expect the results of `.get('planets')` to be an array. So this is basically doing:

```js
state.planets.filter(planet => planet.id === state.meta.selected)
```

Correct? And the main gain being the memoization if you call it multiple times? Writing the query natively feels clearer... is there a way to make this feel as clear as writing it manually?

nhunzaker: I hate the name `join` too... Basically, query functions can specify dependencies. The query won't run again if those dependencies haven't changed.

SoI want a way to filter down from another key while declaring that the query depends on the `meta.selected` field. Another option could be to allow `get` to fetch more than one item:

```javascript
q.get('planets', 'meta.selected')
  .tap((planets, selected) => planets.filter(p => p.id === selected)
```

So how do we make that more concise?

PR #230, opened 2017-02-24
Presenter updates

nhunzaker:
Whew! I finally wrapped this up! This implements the new updates to Presenters:

```javascript
class SolarSystem extends Presenter {

  setup () {
    // runs before `this.getModel` has executed. Useful for setting up a repo fork
  }

  ready () {
    // runs after setup, and the first model calculation
    console.log(this.model) // [{ id: "mercury" }, { id : "venus" } ...]
  }

  getModel () {
    return {
      planets: state => state.planets
    }
  }

  render () {
    const { planets } = this.model
    return <PlanetsView planets={planets} />
  }
}
```

- Presenters that do not implement a model do not listen to repo changes.
- `view` always calls React.createElement
- `render` works as an other React component
- Presenters have access to the current calculated model via `this.model`
- The `update` hook has access to the latest model
- Added `ready`, a hook that fires after setup that has access to the most recent model.
- `getModel` can no longer return a function

Hopefully this makes Presenters feel a _bit_ less magical, and more idiomatic to React.

### DIFF
--- a/docs/api/presenter.md
+++ b/docs/api/presenter.md
@@ -8,9 +8,11 @@
 ## Overview
 
 The Presenter add-on makes it easier to keep application logic high
-within a component tree. It is designed specifically to extract and
-compute properties coming from a Microcosm instance and efficiently
-send them down as `props` to child "passive view" React components.
+within a component tree. It subscribes to state changes via a
+`getModel` method, designed specifically to extract and compute
+properties coming from a Microcosm instance. When state changes, model
+keys are efficiently sent down as props to child “passive view” React
+components.
 
 Presenters also make it easy for components deep within a component
 tree to communicate without passing a long chain of props. The
--- a/docs/api/presenter.md
+++ b/docs/api/presenter.md
@@ -199,7 +199,9 @@ changes.
 
 ### `view`
 
-If a Presenter has a `view` property, it create the associated component instead of calling `render`. The `view` component is given the latest model data:
+If a Presenter has a `view` property, it creates the associated component
+instead of calling `render`. The `view` component is given the latest model
+data:
 
 ```javascript
 function Message ({ message }) {
--- a/docs/api/presenter.md
+++ b/docs/api/presenter.md
@@ -37,14 +37,16 @@ repo.patch({ planets: [ 'Mercury', 'Venus', 'Earth' ]})
 
 class PlanetsPresenter extends Presenter {
 
-  model (props, state) {
+  getModel (props, state) {
     return {
       planets: data => data.planets
     }
   }
 
-  view ({ planets }) {
-    return <p>{ planets.join(', ') }</p>
+  render () {
+    const { planets } = this.model
+
+    return <p>{planets.join(', ')}</p>
   }
 
 }
@@ -56,24 +58,6 @@ In the example above, the `PlanetsPresenter` will extract a list of
 planets from the Microcosm instance provided to it via the `repo`
 prop. This is available as state, which the Presenter can send into a
 child component.
-
-### Listening to all state changes
-
-While we do not recommend it for large repos, some times it's simply easier
-to subscribe to all repo changes. `model` can also return a function, which
-will be called with state:
-
-```javascript
-class PlanetsPresenter extends Presenter {
-  model () {
-    return data => data
-  }
-
-  view ({ planets }) {
-    return <p>{ planets.join(', ') }</p>
-  }
-}
-```
 
 ## Receiving Actions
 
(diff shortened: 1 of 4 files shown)


docs/api/presenter.md
```diff
@@ -203,47 +190,46 @@ class PlanetPresenter extends Presenter {
...
+`getModel` assigns a `model` property to the presenter, similarly to `props` or
+`state`. It is recalculated whenever the Presenter's `props` or `state`
+changes, and functions returned from model keys are invoked every time the repo
+changes.
 
-### `view(model)`
+### `view`
 
-A special render method that is given the current result of `model`.
+If a Presenter has a `view` property, it create the associated component instead of calling `render`. The `view` component is given the latest model data:
```
greypants: ...**creates * the associated component...

docs/api/presenter.md
```diff
@@ -203,47 +190,46 @@ class PlanetPresenter extends Presenter {
...
-class Example extends Presenter {
+class ShowPlanet extends Presenter {
+  getModel (props) {
+    return {
+      planet: state => state.planets.find(p => p.id === props.id)
+    }
+  }
   get view {
-    return MyView
+    return this.model.planet ? PlanetView : MissingView
```
greypants: nice!

greypants:
> Presenters that do not implement a model do not listen to repo changes.

I think this is important to call out in the documentation. It could also be clearer in the [overview of `Presenters`](http://code.viget.com/microcosm/api/presenter.html#overview) that one of their primary purposes **IS** to listen to repo changes and re-render, and that implementing a `getModel` method is necessary to do this.

I'm sure there is one, but what's a use case for a Presenter *without* a `getModel` method?

nhunzaker:
@greypants To perform general prep work and to catch intents to perform general, UI related actions. For example, on RA we have a presenter that acts as a container for multiple views that display charts:

```javascript
class StrategiesPresenter extends Presenter {

  setup (repo, { location }) {
    repo.push(getStrategies).onDone(() => {
      this.fetch(repo, location.query)
    })
  }

  update (repo, { location }) {
    if (location.query.selected !== this.props.location.query.selected ||
        location.query.region !== this.props.location.query.region) {
      this.fetch(repo, location.query)
    }
  }

  fetch (repo, { selected }) {
    const request = repo.prepare(getStrategySeries)
    const {strategies} = repo.state

    let picked = strategies.filter(s => s.Region === repo.state.ui.region.value)
                           .filter(s => within(selected, s.Key))

    picked.forEach(s => request(s.ID))
  }

}
```

nhunzaker:
@greypants We removed blanket passing of state from `model` because it's a dangerous default. It leads to a bunch of recursive re-renders across your UI because the Presenter wants to know _everything_.

greypants:
Re-wording the overview could be something along the lines of:

> The Presenter add-on makes it easier to keep application logic high within a component tree. It subscribes to state changes via a `getModel` method, designed specifically to extract and compute properties coming from a Microcosm instance. When state changes, `model` keys are efficiently sent down as props to child “passive view” React components.

nhunzaker:
@greypants I like that rewording a lot. FWIW, that presenter I shared is the specific reason I want to add `getModel` and the `ready` callback. It could be so much cleaner if had access to the model in `update`, which we'll have soon:

```javascript
class StrategiesPresenter extends Presenter {

  getModel (props) {
    let selected = props.location.query.selected
    
    return {
      missing:  state => strategies.filter(s => s.Region === state.ui.region.value)
                                   .filter(s => within(selected, s.Key))
                                   .filter(s => !s.loading || s.dataPoints)
                                   .map(s => s.id)
    }
  }

  ready (repo) {
    repo.push(getStrategies)
    this.fetchMissingData()
  }

  update () {
    this.fetchMissingData()
  }

  fetchMissingData () {
    let fetch = this.repo.prepare(getDataPoints)
    this.model.missing.forEach(fetch)
  }

}
```

Once we land on an API for queries, that could eventually become:

```javascript
class StrategiesPresenter extends Presenter {

  getModel (props, state, select) {
    let selected = props.location.query.selected

    let isSelected = select('strategies').join('region', 'ui.region.value').within('id', selected)
    let isMissing = isSelected.where({ loading: false, dataPoints: null }).pluck('id')

    return {
      missing: isMissing
    }
  }

  ready (repo) {
    repo.push(getStrategies)
    this.fetchMissingData()
  }

  update () {
    this.fetchMissingData()
  }

  fetchMissingData () {
    let fetch = this.repo.prepare(getDataPoints)
    this.model.missing.forEach(fetch)
  }

}
```

PR #233, opened 2017-02-27
General utility improvements, add update utility helper

nhunzaker:
Updates utility methods to avoid recursion, offering a small perf boost. Also allows keys to be set as dot notation paths like:

```javascript
set(obj, 'a.b.c', value)
```

It also adds an `update` helper. Often times, we want to modify data in place. This commit adds an `update` helper that works like a combination of `get` and `set`:

```javascript
import { update } from 'microcosm'

let votes = { yay: 0, nay: 0 }
let next = update(votes, 'yay', n => n + 1)

console.log(next) // { yay: 1, nay: 0 }
```

`update` also excepts a fallback value. If the provided target is missing, the keypath is missing, the processor function will be called with the fallback value:

```javascript
import { update } from 'microcosm'

let votes = { yay: 0, nay: 0 }
let next = update(votes, 'undecided', n => n + 1, 0)

console.log(next) // { yay: 0, nay: 0, undecided: 1 }
```

### DIFF
--- a/docs/api/data-utilities.md
+++ b/docs/api/data-utilities.md
@@ -79,9 +79,8 @@ let next = update(votes, 'yay', n => n + 1)
 console.log(next) // { yay: 1, nay: 0 }
 ```
 
-`update` also excepts a fallback value. If the provided target is missing,
-the keypath is missing, the processor function will be called with the fallback
-value:
+`update` also accepts a fallback value. If the provided keypath is missing, the
+processor function will be called with the fallback value:
 
 ```javascript
 import { update } from 'microcosm'
--- a/docs/api/data-utilities.md
+++ b/docs/api/data-utilities.md
@@ -65,6 +65,33 @@ let next = set(state, ['planets', 'earth', 'color'], blue)
 console.log(next === subject) // true
 ```
 
+## update(target, key, value, processor, fallback)
+
+Extract a value from a keypath, then call a function on that value and assign
+the result at the same point. This is useful for modifying values in-place:
+
+```javascript
+import { update } from 'microcosm'
+
+let votes = { yay: 0, nay: 0 }
+let next = update(votes, 'yay', n => n + 1)
+
+console.log(next) // { yay: 1, nay: 0 }
+```
+
+`update` also excepts a fallback value. If the provided target is missing,
+the keypath is missing, the processor function will be called with the fallback
+value:
+
+```javascript
+import { update } from 'microcosm'
+
+let votes = { yay: 0, nay: 0 }
+let next = update(votes, 'undecided', n => n + 1, 0)
+
+console.log(next) // { yay: 0, nay: 0, undecided: 1 }
+```
+
 ## merge(...objects)
 
 Non-destructively merge together the keys of any number of objects,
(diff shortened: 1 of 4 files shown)


docs/api/data-utilities.md
```diff
@@ -65,6 +65,33 @@ let next = set(state, ['planets', 'earth', 'color'], blue)
...
+```javascript
+import { update } from 'microcosm'
+
+let votes = { yay: 0, nay: 0 }
+let next = update(votes, 'yay', n => n + 1)
+
+console.log(next) // { yay: 1, nay: 0 }
+```
+
+`update` also excepts a fallback value. If the provided target is missing,
```
mackermedia: sp: "accepts"

mackermedia: `If the provided target is missing, the keypath is missing, the processor function ...`
Should this just be `If the provided keypath is missing on the provided target, the process function ...` ?

src/utils.js
```diff
@@ -96,49 +99,59 @@ export function get (object, path, fallback) {
...
- * object.
+ * Non-destructively assign a value to a provided object at a given key. If the
+ * value is the same, don't do anything. Otherwise return a new object.
  */
-export function set (object, path, value) {
-  if (value === undefined || get(object, path) === value) {
-    return object
-  }
-
+export function set (object, path, goal) {
```
mackermedia: Is `goal` the same as `fallbackValue` ?

mackermedia: Oh, nvm. Although I think I prefer the original `value` over `goal`.

src/utils.js
```diff
@@ -96,49 +99,59 @@ export function get (object, path, fallback) {
...
+  return root
 }
+
+
+/**
+ * A helper combination of get and set
+ */
+ export function update (state, path, fn, fallback) {
+   let last = get(state, path, fallback)
+   let next = fn(last)
```
mackermedia: What if you wanted to use `update` similar to `get` and provide a default value instead of a default function? Will `fn(last)` if it's a `'defaultString'` blow up?

test/utils.test.js
```diff
@@ -32,6 +34,25 @@ describe('clone', function () {
...
+
+  it('does not clone booleans', function () {
+    expect(clone(true)).toBe(true)
+  })
+
+  it('does not clone null', function () {
+    expect(clone(null)).toBe(null)
+  })
+
+  it('does not undefined null', function () {
```
mackermedia: `does not clone undefined`

mackermedia:
Couple small things. Neat-o! 👍

docs/api/data-utilities.md
nhunzaker: .... Jeez. Thank you.

nhunzaker: yes yes yes

src/utils.js
nhunzaker: Okay. It will be so.

src/utils.js
nhunzaker: Hmm. that's a good point. So you're saying use the fallback value as is, instead of calling `fn`on it?

test/utils.test.js
nhunzaker: Thank ya. Man....

nhunzaker: Maybe I should do this work later in the mornings...

src/utils.js
mackermedia: Yeah, like allowing it to work with either a static value or a `fn`

Issue #226, opened 2017-02-22
Dealing with lots of in progress actions


nhunzaker:
Right now, our action reconciliation process works like this:

```
1-done -> 2-open -> 3-open -> 4-open -> 5-open -> 6-done
```

When ever an action changes, we roll back to the current working root. In the example above, a change an action will result in the following sequence of events:

1. Roll back to last "stable" state. This is action 1 in the diagram.
2. Roll forward, reprocessing all future actions (3, 4, 5, 6)
3. Release the new state

This can be better. Currently, we only maintain a cache of the state at the stable root (action 1). However that means that if actions resolve in reverse order, we operate on the summation of the size of the tree. 

1. action 5 finishes: Process 2,3,4,5,6
2. action 4 finishes: Process 2,3,4,5,6
3. action 3 finishes: Process 2,3,4,5,6
4. action 2 finishes: Process 2,3,4,5,6
5. Mark 6 as the next root

So not ideal. The reconciliation process is extremely fast, but there is still waste. We could avoid this by tracking state for outstanding actions and progressively rolling forward, like:

1. action 5 finishes: Process 5,6
2. action 4 finishes: Process 4,5,6
3. action 3 finishes: Process 3,4,5,6
4. action 2 finishes: Process 2,3,4,5,6
5. Mark 6 as the next root

PR #236, opened 2017-03-06
History improvements

nhunzaker:
I'm working on an undo/redo feature on a project and I already bumped into issues where we over-eagerly rollback actions. So I fixed it. This commit changes the action resolution algorithm so that history only needs to roll back to the point of change, preventing excessive Domain handler modification.

For example, assuming the following tree:

```
1-done -> 2-open -> 3-open -> 4-open -> 5-open -> 6-done
```

**Before:**

1. action 5 finishes: Process 2,3,4,5,6
2. action 4 finishes: Process 2,3,4,5,6
3. action 3 finishes: Process 2,3,4,5,6
4. action 2 finishes: Process 2,3,4,5,6
5. Mark 6 as the next root

**After:**

1. action 5 finishes: Process 5,6
2. action 4 finishes: Process 4,5,6
3. action 3 finishes: Process 3,4,5,6
4. action 2 finishes: Process 2,3,4,5,6
5. Mark 6 as the next root

This also got rid of the need for a bunch of null checks, which makes things faster and cleaner.

--- 

Fix for https://github.com/vigetlabs/microcosm/issues/226

### DIFF
--- a/src/action.js
+++ b/src/action.js
@@ -15,7 +15,10 @@ export default function Action (command, history, status) {
 
   this.id = this.history.getId()
   this.command = tag(command)
-  this.status = status || 'inactive'
+
+  if (status) {
+    this.status = status
+  }
 }
 
 inherit(Action, Emitter, {
--- a/src/action.js
+++ b/src/action.js
@@ -5,18 +5,20 @@ import { ACTION_STATES } from './constants'
 import { inherit } from './utils'
 
 /**
- * Actions encapsulate the process of resolving an action creator. Create an
- * action using `Microcosm::push`:
+ * Actions encapsulate the process of resolving an action
+ * creator. Create an action using `Microcosm::push`:
  */
-export default function Action (command, history) {
+export default function Action (command, history, status) {
   Emitter.call(this)
 
-  this.command = tag(command)
   this.history = history || new History()
+
+  this.id = this.history.getId()
+  this.command = tag(command)
+  this.status = status || 'inactive'
 }
 
 inherit(Action, Emitter, {
-  type       : null,
   status     : 'inactive',
   payload    : undefined,
   disabled   : false,
@@ -33,7 +35,7 @@ inherit(Action, Emitter, {
   toggle () {
     this.disabled = !this.disabled
 
-    this.history.invalidate()
+    this.history.reconcile(this)
 
     return this
   },
@@ -44,7 +46,6 @@ inherit(Action, Emitter, {
       this.onError(reject)
     }).then(pass, fail)
   }
-
 })
 
 /**
--- a/src/archive.js
+++ b/src/archive.js
@@ -1,0 +1,25 @@
+/**
+ * Keep track of prior action states according to an action's id
+ */
+
+export default function Archive () {
+  this.pool = {}
+}
+
+Archive.prototype = {
+  get (action) {
+    return this.pool[action.id]
+  },
+
+  has (action) {
+    return this.pool.hasOwnProperty(action.id)
+  },
+
+  set (action, state) {
+    this.pool[action.id] = state
+  },
+
+  remove (action) {
+    delete this.pool[action.id]
+  }
+}
--- a/src/history.js
+++ b/src/history.js
@@ -1,28 +1,36 @@
-import Action from './action'
-
 /**
  * The central tree data structure that is used to calculate state for
  * a Microcosm. Each node in the tree represents an action. Branches
  * are changes over time.
  */
-export default function History (limit=0) {
+
+import Action from './action'
+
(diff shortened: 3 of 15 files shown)


src/action.js
```diff
@@ -5,18 +5,20 @@ import { ACTION_STATES } from './constants'
...
+ * creator. Create an action using `Microcosm::push`:
  */
-export default function Action (command, history) {
+export default function Action (command, history, status) {
   Emitter.call(this)
 
-  this.command = tag(command)
   this.history = history || new History()
+
+  this.id = this.history.getId()
```
nhunzaker: Each action gets a unique ID, provisioned by its history.

src/archive.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+
+  set (action, state) {
+    this.pool[action.id] = state
+  },
+
+  remove (action) {
+    delete this.pool[action.id]
+  }
+}
```
nhunzaker: This is a simple `Map` of repo snapshots for a given action.

src/history.js
```diff
@@ -1,28 +1,36 @@
...
+
+export default function History (limit) {
+  this.ids = 0
+  this.size = 0
+  this.limit = Math.max(1, limit || 1)
   this.repos = []
-  this.limit = limit
+
+  this.append('_root', 'resolve')
 }
```
nhunzaker: History now _always_ has a root node. At the start, we append a completed placeholder.

src/history.js
```diff
@@ -1,28 +1,36 @@
...
+import Action from './action'
+
+export default function History (limit) {
+  this.ids = 0
+  this.size = 0
+  this.limit = Math.max(1, limit || 1)
   this.repos = []
-  this.limit = limit
+
+  this.append('_root').resolve()
```
nhunzaker: Each history starts with a resolved root node. This avoids quite a few null checks.

src/history.js
```diff
@@ -74,76 +81,46 @@ History.prototype = {
...
+    while (this.size > this.limit && action.disposable) {
       this.size -= 1
 
-      if (this.size <= 0) {
-        this.root = this.head = this.focus = null
-      } else {
-        this.root = this.root.next
-        this.root.parent = null
+      if (action.parent) {
+        this.invoke('clean', action.parent)
```
nhunzaker: I think there's a next step here where we don't dispatch a `clean` command until we've hit the new root, but I couldn't quite figure it out.

src/microcosm.js
```diff
@@ -82,113 +67,75 @@ inherit(Microcosm, Emitter, {
...
-          next = handler.call(domain, get(state, key))
-          break;
-        case 2:
-        default:
-          next = handler.call(domain, get(state, key), action.payload)
-      }
+      var { key, domain, handler } = handlers[i]
+
+      var last = get(state, key)
+      var next = handler.call(domain, last, action.payload)
```
nhunzaker: This was a perf opt, but it throws off Jest and I don't want to hit that on a project.

src/realm.js
```diff
@@ -31,7 +32,7 @@ Realm.prototype = {
   add (key, config, options) {
     let domain = createOrClone(config, options, this.repo)
 
-    this.domains.push([key, domain])
+    this.domains.push([castPath(key), domain])
```
nhunzaker: This is a perf opt. We avoid re-casting string keys every time we dispatch. Pretty nice bump.

src/realm.js
```diff
@@ -67,7 +68,7 @@ Realm.prototype = {
       }
 
       return memo
-    }, seed || {})
+    }, seed)
```
nhunzaker: No need to default this anymore :)

src/utils.js
```diff
@@ -1,13 +1,12 @@
...
+  if (value == null) {
+    return []
+  } else if (Array.isArray(value)) {
     return value
-  } else if (value == null) {
-    return EMPTY_ARRAY
   }
 
-  return typeof value === 'string' ? value.split('.') : [value]
+  return typeof value === 'string' ? value.split(SPLITTER) : [value]
```
nhunzaker: Perf opt. passing in an existing string is about 15% faster

src/action.js
```diff
@@ -5,18 +5,20 @@ import { ACTION_STATES } from './constants'
...
+  this.command = tag(command)
+  this.status = status || 'inactive'
```
greypants: worth storing status strings in constants?

nhunzaker: Good call.

nhunzaker: Actually, this is already set as the default. I should just do `if(status) this.status = status`

PR #239, opened 2017-03-06
Coverage fix

nhunzaker:
cc @leobauza


leobauza:
This makes the documentation accurate again

PR #244, opened 2017-03-09
Invoke should not run on removed repos

nhunzaker:
This commit fixes a bug where Microcosm would raise an exception if a
fork was removed from history as the result of a parent changing. It
also adds an assertion to make it clearer during testing.

Sorry, @dce :(

### DIFF
--- a/test/unit/history/invoke.test.js
+++ b/test/unit/history/invoke.test.js
@@ -1,0 +1,23 @@
+import History from '../../../src/history'
+import Microcosm from '../../../src/microcosm'
+
+describe('History::invoke', function() {
+
+  it('it does invoke a repo that no longer exists', function() {
+    let parent = new Microcosm()
+    let child = parent.fork()
+
+    parent.on('change', () => child.teardown())
+
+    jest.spyOn(child, 'release')
+
+    parent.addDomain('test', {
+      getInitialState() {
+        return false
+      }
+    })
+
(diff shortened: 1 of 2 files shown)


test/unit/history/invoke.test.js
```diff
@@ -0,0 +1,23 @@
...
+    jest.spyOn(child, 'release')
+
+    parent.addDomain('test', {
+      getInitialState() {
+        return false
+      }
+    })
+
+    expect(child.release).not.toHaveBeenCalled()
+  })
```
mackermedia: I'm not super familiar with the lifecycle hooks, but reading this test it isn't obvious to me why the child goes away (child.teardown()?) triggered by adding a domain on the parent? And if so, why would `child.release` not be called? What is `child.release` ?

mackermedia:
clarifying question, but 👍

test/unit/history/invoke.test.js
nhunzaker: Still waiting for some details on that, however I was investigating and found this edge case, so I went ahead and stamped it out.

So we'll see :)

nhunzaker: As for `release`, `release` is an internal Microcosm method for sending out a change event. It fires a change, then it invokes Effects with the related action.

This allows a tree of microcosms (parent + forks) to all resolve an action, then handle change subscriptions when every microcosm is up to date.

PR #252, opened 2017-03-10
Properly release state changes to children

nhunzaker:
This commit prevents a bug where children would not receive the latest
updates when their parents emitted a change event.

I believe this was introduced in 12.1.0.

### DIFF
--- a/src/microcosm.js
+++ b/src/microcosm.js
@@ -44,6 +44,11 @@ function Microcosm (options, state, deserialize)  {
   // Microcosm is now ready. Call the setup lifecycle method
   this.setup(options)
 
+  // Track a dirty flag, this is useful so that we can reconcile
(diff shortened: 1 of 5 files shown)


src/microcosm.js
```diff
@@ -126,12 +131,17 @@ inherit(Microcosm, Emitter, {
...
-      this._emit('change', next)
+    this.dirty = next !== this.state
+    this.state = next
+  },
+
+  release (action) {
+    if (this.dirty) {
+      this.dirty = false
+      this._emit('change', this.state)
     }
```
nhunzaker: I had this before, but removed it when I switched to the new reconciliation approach :-/

mackermedia:
👍

nhunzaker:
Thanks ya

Issue #250, opened 2017-03-10
Docs gaps and common pitfalls


leobauza:
@nhunzaker and I were talking earlier about common problems (maybe we need a common gotchas/mistakes/whatever section) and about internal documentation. I wanted to make sure that this didn't get lost:

Generally, though, there's really only a few frequent problems:

1. People forget to close out actions
2. People don't register actions correctly (typos)
3. People don't register action handlers correctly (typos)

If anyone has any common problems, gaps in documentation, or hidden documentation that we aren't exposing at a top level (ie. the site, the top level README) then comment here.

cwmanning:
https://github.com/vigetlabs/microcosm/blob/master/docs/api/microcosm.md#addeffectconfig-options has a code example that is adding a domain.

cwmanning:
The difference between `patch` and `reset` is not clear to me. Is it possible to show an example input/output for those to illustrate the difference?

nhunzaker:
We just hit another one. `withSend` wraps components, which can be confusing when using [Enzyme's](https://github.com/airbnb/enzyme) shallow rendering.

PR #253, opened 2017-03-10
Fix effect docs in microcosm api documentation

nhunzaker:
Thanks for catching this, @cwmanning


leobauza:
Noting that this addresses parts of #250

docs/api/microcosm.md
```diff
@@ -104,20 +104,38 @@ manage the provided `key`. Whenever this domain responds to an action,
 it will be provided the current state for that particular key.
 
 `options` passed as the third argument are sent into a domain's
-`setup` method and, if using a class, the constructor.
+`setup` method and, if using a class, the constructor is instantiated
+with the provided options and associated repo.
```
leobauza: Why is there no code sample for `addDomain` like there is for `addEffect` ?

nhunzaker: There is no reason. I'll add an issue for this. We also need to better describe how you can create domains and effects (simple objects vs classes)

nhunzaker:
@leobauza I merged this, I'll follow up with another PR for the domains stuff.

Issue #247, opened 2017-03-10
Update site/docs


leobauza:
Main things that need to happen:

- [x] Update the `README.md` with better instructions to run locally and publish
- [x] Add a "Contributing" section
- [ ] ~~Add a "Roadmap" section~~ (see #258)
- [x] Add an overview for each subsection (ie. Introduction, Api, Addons, etc)
- [ ] Review all language for clarity and accuracy

@nhunzaker would like to get your thoughts on what I am proposing to add (ie. contributing section, roadmap, overviews).

nhunzaker:
I kind of did:

> Add an overview for each subsection (ie. Introduction, Api, Addons, etc)

What do you think?

leobauza:
Yeah, the only thing missing is an explanation of the difference between API and Addon. That is that the Addons are specific to react and API doesn't depend on a framework. But yeah your overviews pretty much covered what we needed. The quickstart or architecture section can make the division clearer when introducing concepts.

leobauza:
Closing and moving to the road to JSDoc #316

PR #255, opened 2017-03-13
[#247] Update site README with more detailed instructions

leobauza:
Addresses 1 item of #247 

This PR adds more details on the contents of the `site` folder, more detailed instructions, and some background about technologies being used.


site/README.md
```diff
@@ -23,11 +51,17 @@ cd site
...
 ```
 
+The html files will be generated (or updated) into `./public` and a server will be started running on `localhost:4000`. Point your browser there to see the site. Changes to files in `./src` or any `.md` file will trigger a live reload on the site (no need to hit refresh).
+
+> To see what is going on under the hood look at the `Makefile` and at the `./scripts/serve` script. If you are not familiar with browser-sync check out [their website](https://www.browsersync.io/) for details.
+
 ## Publish
 
-From the project root:
+Publishing to `http://code.viget.com/microcosm` can only be done by core contributors with merging privileges. If this is you then publishing is straightforward, from the project root:
```
leobauza: Are there other reasons to use the publish script? Contributors shouldn't submit PR's from the `gh-pages` branch right? We will generate the site ourselves when it's ready.

nhunzaker: That is correct. I don't think there is a reason for a non-managing contributor to use the publish script.

PR #256, opened 2017-03-13
Removing an action from history

nhunzaker:
This commit adds a new private API for removing an action from the history tree. Long term, this is part of an effort to expose a public history management API for undo/redo features.

Some of these operations feel a little too mechanical. The next step will be to see how we can make things a little more elegant. Still, this implements the basic behavior in a way that doesn't effect existing public APIs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 12.2.0 (Beta)
+## 12.2.0
 
 - Added method for removing actions from history, history `append`
   event.


src/lifecycle.js
```diff
@@ -16,14 +16,18 @@ function sandbox (data, deserialize) {
...
 
-export const BIRTH = '$birth'
+export const BIRTH = function $birth () {
+  console.assert(false, 'Birth lifecycle method should never be invoked directly.')
+}
 
-export const START = '$start'
+export const START = function $start () {
+  console.assert(false, 'Start lifecycle method should never be invoked directly.')
+}
```
nhunzaker: Writing these as functions gives me a referenceable name from the function. Long term, I would like to see how I could convert a string action like "$start" into an identity function that reports a name of the same string value.

solomonhawk:
Lgtm :+1:

nhunzaker:
Newest commit resolves conflicts with master. I forgot we shipped the action fix after I released this PR.

PR #257, opened 2017-03-13
Fix nested snapshots

nhunzaker:
We create a snapshot of the state for every action so that we call rollback to that state when resolving changes to actions. This commit fixes a bug where actions pushed inside of other actions would cause a rollback to a snapshot that hadn't been saved yet, effectively reverting to initial state.

The original code that introduced the bug:

```javascript
export function spamCheck () {
  return function (action, repo) {
    repo.push(focusBlock, null)

    // Operates on action, pushing more actions
    Api().spamCheck(action, repo)
  }
}
```

### DIFF
--- a/src/history.js
+++ b/src/history.js
@@ -84,6 +84,8 @@ History.prototype = {
     this.head = action
     this.size += 1
 
+    this.invoke('createInitialSnapshot', action)
+
     return this.head
   },
 
@@ -105,7 +107,7 @@ History.prototype = {
 
     this.archive()
 
-    this.invoke('update')
+    this.invoke('prepareRelease')
 
     this.invoke('release', action)
   },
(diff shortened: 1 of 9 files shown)


src/history.js
```diff
@@ -116,7 +118,7 @@ History.prototype = {
 
     while (size > this.limit && root.disposable) {
       size -= 1
-      this.invoke('clean', root.parent)
+      this.invoke('removeSnapshot', root.parent)
       root = root.next
     }
```
nhunzaker: I renamed these methods to be a little clearer for what's happening.

cwmanning:
👍 . sounds sort of related to an `onprogress` issue I ran into earlier this year, will try it out later.

Issue #260, opened 2017-03-15
new props don't cause Presenters to re-render


greypants:
**This issue popped up when upgrading from `11.6.0` to `12.2.0`**

I noticed this issue when components stopped swapping out on route changes. Simplified, the issue looks like this:

```js
class Application extends Presenter {
  render() {
    <div className="some-layout">
      { this.props.children }
    </div>
  }
}
```

Routes:
```js
<Route path="/" component={Application}>
  <Route path="foo" component={Foo} />
  <Route path="bar" component={Bar} />
</Route>
```

Changing from `/foo` to `/bar`, does not cause a re-render, and the component is not swapped out.

Simply adding `children` to the model seems to force the Presenter to update when the prop changes, and works around the issue.

```js
class Application extends Presenter {
  getModel (props) {
    return {
      children: () => props.children
    }
  },

  render () {
    <div className="some-layout">
      { this.props.children }
    </div>
  }
}
```

But as I understand it, prop changes should cause `render` to be called again.

nhunzaker:
This should be fixed in 12.2.1, currently in release candidate. Will close after we publish to stable.

nhunzaker:
Fixed: https://github.com/vigetlabs/microcosm/releases/tag/v12.2.1

Issue #259, opened 2017-03-15
repo.addDomain reseting other domain state


greypants:
**This issue popped up when upgrading from `11.6.0` to `12.2.0`**

Domain state is getting reset to the result of `getInitialState` for some reason after calling `addDomain`:

```js
export default class LicenseManager extends Presenter {
  setup(repo, props) {
    console.log('LicenseManager BEFORE addDomain', repo.state.user)
    repo.addDomain('licenses', LicensesDomain)
    console.log('LicenseManager AFTER addDomain', repo.state.user)
  }
  …
```
Logs the following:

![logs](https://cloud.githubusercontent.com/assets/647549/23967476/710daa7c-0995-11e7-9b6c-6774d08478ad.png)

The second log is the initial state of the `user` domain.

This `Presenter` is nested under two other parent presenters.

nhunzaker:
This should be fixed in 12.2.1, currently in release candidate. Will close after we publish to stable.

nhunzaker:
Fixed https://github.com/vigetlabs/microcosm/releases/tag/v12.2.1

PR #261, opened 2017-03-16
Fix patch domains

nhunzaker:
Available for testing by installing `microcosm@12.2.1-alpha.0`

- adding domains does not reset state
- patching parents with keys owned by children does not patch children
- patch does not reset state in forks
- new children props re-render presenters
- presenter.send method is autobound, allowing it work when passed to children
---

Related issues:
https://github.com/vigetlabs/microcosm/issues/260
https://github.com/vigetlabs/microcosm/issues/259

### DIFF
--- a/src/addons/presenter.js
+++ b/src/addons/presenter.js
@@ -16,6 +16,9 @@ function Presenter (props, context) {
   } else {
     this.defaultRender = passChildren
   }
+
+  // Autobind send so that context is maintained when passing send to children
+  this.send = this.send.bind(this)
 }
 
 inherit(Presenter, PureComponent, {
@@ -89,7 +92,8 @@ inherit(Presenter, PureComponent, {
     return (
       createElement(PresenterMediator, {
         presenter   : this,
-        parentState : this.state
+        parentState : this.state,
+        parentProps : this.props
       })
     )
   }
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -12,7 +12,12 @@ function sandbox (data, deserialize) {
       }
     }
 
-    action.resolve(payload)
+    // Strip out keys not managed by this repo. This prevents
+    // children from accidentally having their keys reset by
+    // parents.
+    let sanitary = repo.realm.prune({}, payload)
+
+    action.resolve(sanitary)
   }
 }
 
@@ -30,4 +35,8 @@ export const BIRTH = function $birth () {
 
 export const START = function $start () {
   console.assert(false, 'Start lifecycle method should never be invoked directly.')
+}
+
+export const ADD_DOMAIN = function $addDomain (domain) {
+  return domain
 }
--- a/test/unit/microcosm/patch.test.js
+++ b/test/unit/microcosm/patch.test.js
@@ -98,6 +98,72 @@ describe('Microcosm::patch', function () {
       // the new state produced by a child.
       expect(child.state.count).toEqual(4)
     })
+
+    it('does not strip away parent inherited state', function () {
+      const parent = new Microcosm()
+      const child = parent.fork()
+
+      parent.addDomain('top', {
+        getInitialState() {
+          return false
+        }
+      })
+
+      child.addDomain('bottom', {
+        getInitialState() {
+          return false
+        }
+      })
+
+      child.patch({ bottom: true })
+
+      expect(child).toHaveState('top', false)
+      expect(child).toHaveState('bottom', true)
+    })
+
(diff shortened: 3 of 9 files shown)


src/addons/presenter.js
```diff
@@ -89,7 +92,8 @@ inherit(Presenter, PureComponent, {
     return (
       createElement(PresenterMediator, {
         presenter   : this,
-        parentState : this.state
+        parentState : this.state,
+        parentProps : this.props
```
nhunzaker: This addition here fixes the `children` issue.

src/lifecycle.js
```diff
@@ -12,7 +12,12 @@ function sandbox (data, deserialize) {
       }
     }
 
-    action.resolve(payload)
+    // Strip out keys not managed by this repo. This prevents
+    // children from accidentally having their keys reset by
+    // parents.
+    let sanitary = repo.realm.prune({}, payload)
+
+    action.resolve(sanitary)
```
nhunzaker: This prevents parents from patching state of children.

test/unit/microcosm/patch.test.js
```diff
@@ -98,6 +98,72 @@ describe('Microcosm::patch', function () {
...
+        }
+      })
+
+      child.addDomain('bottom', {
+        getInitialState() {
+          return false
+        }
+      })
+
+      child.patch({ bottom: true })
```
cwmanning: question about `patch`: in this example, if the domain state is an object, what's the expected output?

```js
child.addDomain('example', {
  getInitialState() { return { id: 1, name: 'Billy' } }
}

child.patch({ example: { name: 'Ben' } })
```

I think I would expect `example` to be `{ id: 1, name: 'Ben' }` but wanted to check and see if there's a test for that already.

src/addons/presenter.js
greypants: That's where I figured the change would go, but I was having a hard time figuring out how `parentProps` and `parentState` actually get used and passed down. They get passed to a mediator component as props... then what?

cwmanning:
👍  thanks for the work on these.

test/unit/microcosm/patch.test.js
```diff
@@ -98,6 +98,72 @@ describe('Microcosm::patch', function () {
...
+        getInitialState() {
+          return false
+        }
+      })
+
+      parent.patch({ bottom: true })
+
+      expect(parent).not.toHaveState('bottom')
+      expect(child).toHaveState('bottom', false)
+    })
```
greypants: Do you need a test that asserts something like

```js
it('merges state', function() {
  repo.addDomain('foo', {
    getInitialState() {
      return {
        a: 'a'
      }
    }
  })
    
  repo.patch({
    foo: {
      b: 'b'
    }
  })

  expect(repo.state.foo.a).toBe('a')
  expect(repo.state.foo.b).toBe('b')
})
```

Also, how do you have `toHaveState`? Is that a Jest add-on?

src/addons/presenter.js
nhunzaker: They get passed down to bust the behavior of `React.PureComponent`. I do not think they are actually used by the mediator.

src/lifecycle.js
nhunzaker: Prune loops over all of the domains, copying over their keys from one object to another. This gets rid of any keys that aren't within what are specified when you call `repo.addDomain`.

The first argument, `{}` here, is the object to copy the values into. It's sort of like a base state.

test/unit/microcosm/patch.test.js
greypants: cc @cwmanning - related to your comment. I had the same question.

nhunzaker: It's an internal Microcosm thing. `addons/jest-matchers`. They make my life easier, and I want to share them with others. I just don't think they're quite ready yet.

nhunzaker: @greypants patch doesn't do a deep merge, so that test would fail.

test/unit/microcosm/patch.test.js
nhunzaker: Ah interesting. Patch doesn't do a deep merge. The primary use case is to seed data. If you need a deep merge, the best (and only, I think) approach is to use an action.

nhunzaker:
Thanks gents.

test/unit/microcosm/patch.test.js
greypants: TMYK 🌈 🌠

PR #263, opened 2017-03-21
Add more formal class for domain and effect managers

nhunzaker:
No external changes. This is just a general refactor to help make contributions easier.

---

This commit adds EffectEngine and DomainEngine classes to more
formally manage Effects and Domains.

Domains and Effects manage their own dispatch processes. Long term, I
could see using `repo.domains.add` and `repo.effects.add` instead of
the similar `repo.addDomain` and `repo.addEffect`. Before we do that,
we should figure out what more public access to `repo.effects` and
`repo.domains` might afford us.

Additionally:

- Fix bug where `castPath` would return `['']` when given `''`, instead of an empty array
- Fixed deoptimization in `get` function that was really slowing things down 
- Added `has` utility. We'll keep this internal for now, I don't want
  to expose too many utilities and cause overlap with lodash.


src/action.js
```diff
@@ -145,27 +143,34 @@ inherit(Action, Emitter, {
...
   /**
    * Create a method to update the action status. For example:
    * action.done({ id: 'earth' })
    */
-  Action.prototype[key] = function (payload) {
+  Action.prototype[status] = function (payload) {
     if (!this.disposable) {
-      this.setStatus(key)
+      this.status = status
+      this.disposable = disposable
```
nhunzaker: There is duplication here with a`action.setStatus`. I need to think more on this... By being more static here, V8 can optimize these functions way harder.

src/emitter.js
```diff
@@ -79,10 +75,6 @@ Emitter.prototype = {
    * @param {*} payload Value to send with callback
    */
   _emit (event, payload) {
-    if (this._events == null) {
-      return this
-    }
-
     let i = 0
     while (i < this._events.length) {
       var cb = this._events[i]
```
nhunzaker: By always having an array, V8 can optimize the event emitter way harder. The increase to memory usage is negligible.

src/utils.js
```diff
@@ -1,9 +1,12 @@
...
 const SPLITTER = '.'
 export function castPath (value) {
-  if (value == null) {
-    return []
-  } else if (Array.isArray(value)) {
+  if (Array.isArray(value)) {
     return value
+  } else if (value === '' || value === null || value === undefined) {
+    return EMPTY
   }
```
nhunzaker: This looks pedantic, but it grants V8 sufficient type information to give this function a pretty high level of optimization. Making this change roughly doubled the speed of the `push` performance benchmark.

src/utils.js
```diff
@@ -55,34 +58,16 @@ export function merge () {
...
 export function inherit (Child, Ancestor, proto) {
-  Child.__proto__ = Ancestor
+  if (Ancestor) {
+    Child.__proto__ = Ancestor
+  }
 
   Child.prototype = merge(Object.create(Ancestor.prototype), {
-    constructor: Child
+    constructor: Child.prototype.constructor
   }, proto)
```
nhunzaker: I changed this to be more inline with how Buble (buble.surge.sh) transpiles classes. I'm starting to wonder if we should just use classes again. They are much nicer to work with. 

Thoughts @greypants @mackermedia ?

PR #264, opened 2017-03-21
Add comparison-tree

nhunzaker:
This commit adds a new data structure for change management. Instead
of listening, generically, to `change`, a consumer can now subscribe
to multiple, fine grained, data subscriptions. This looks like:

```javascript
repo.on('change:users,meta.focused', function (users, focused) {
  let user = users.find(user => user.id === focused)

  // Now do something with that user
})
```

This is accomplished by building a tree of key paths, these key paths
represent nested properties within the repository state object.

It's also fast. Very fast. Current simulations run at an upward limit
of around 20 million subscriptions. I think it could probably get as fast as 40 or 50 million. It kind of stops mattering after around 5.

Demo:
http://probable-pocket.surge.sh


src/microcosm.js
```diff
@@ -121,23 +116,40 @@ inherit(Microcosm, Emitter, {
...
+    switch (event) {
+      case 'change':
+        this.changes.off(meta, callback, scope)
+        break;
+      default:
+        Emitter.prototype.off.apply(this, arguments)
     }
 
-    this.effects.dispatch(action)
+    return this
```
nhunzaker: Switching like this is for backwards compatibility. I don't want to break apps listening to `on('teardown')`, even though I don't believe it is documented.

examples/canvas/app/boot.js
```diff
@@ -0,0 +1,96 @@
...
+  while (n-- > 0) {
+    let x = Math.random() * rows | 0
+    let y = Math.random() * columns | 0
+
+    repo.push(advance, x, y)
+  }
+}
+
+function updateLabel () {
+  let events = rows * columns * writes
```
efatsi: why is this calculation valid? I'm seeing this:
```
71x42 (20px grid)| 494 writes/frame| 1,473,108 events/frame
```
looking at the code it looks like that 494 number is the number of "things" happening per frame, i assume i have a misunderstanding of what an "event" is in microcosm

nhunzaker: Ah event is a bad name. This is, generally, how many keys the system can watch per 16 ms. 

This actually probably needs to go, or some other measure needs to be put here. I added this back when the data binding traversing was linear.

So this is probably more like this can crawl "2,982" cells 494 times per frame.

src/compare-tree/key-path.js
```diff
@@ -0,0 +1,40 @@
+const KEY_DELIMETER  = '.'
+const PATH_DELIMETER = ','
```
efatsi: open to some flexibility on this? I would expect something like
```
const KEY_DELIMETER  = '-'
const PATH_DELIMETER = ','
```
if i'm understanding what these two do, i don't actually see anything splitting on the `KEY_DELIMETER` which i'm confused by

examples/canvas/app/boot.js
efatsi: ah interesting, i came into this assuming that the system could now handle millions of actions per frame, but the number of subscriptions is the concern here. would cranking up that # of writes/frame be compelling?

src/compare-tree/key-path.js
```diff
@@ -0,0 +1,40 @@
...
+
+  if (Array.isArray(value) === false) {
+    paths = `${paths}`.split(PATH_DELIMETER)
+  }
+
+  return paths.map(castPath)
+}
+
+export function getKeyStrings (array) {
+  let sorted = array.slice(0).sort(byKeyLength)
```
efatsi: what does `slice(0)` do here?

nhunzaker: Sorting is mutative in JS (😢 ), so this is just a fast way to copy an array.

src/compare-tree/key-path.js
efatsi: ah, looks like the `'.'` convention is already defined in this project?https://github.com/vigetlabs/microcosm/blob/93568b41a3a6e3cf55474d645eca6d79d9ae64a6/src/utils.js#L4

nhunzaker: It is. Slipped my mind. I tried to keep this stuff pretty separate from the rest of the project, but this should be consolidated.

src/compare-tree/key-path.js
```diff
@@ -0,0 +1,40 @@
...
+    return []
+  }
+
+  let paths = value
+
+  if (Array.isArray(value) === false) {
+    paths = `${paths}`.split(PATH_DELIMETER)
+  }
+
+  return paths.map(castPath)
```
efatsi: curious why `castPath` is necessary, seems like you're splitting on the PATH_DELIMETER (`.`), and `castPath` does the same thing. is this covering a case where you pass in an array of items where each item could contain multiple paths (hoping not because that would be a nested array which doesn't sound nice here)

nhunzaker: `PATH_DELIMETER` is a comma. 

This function turns a list of key paths, like `users,query.selected` into `[ ['users'], ['query', 'selected' ] ]`

That way, you can set up a subscription that listens to both key paths at once, executing a callback with the values of both as parameters, like `function (users, selected) {}`

efatsi: 👍  i think i've been switching what key and path mean in my head for the last 10 minutes, should be settled now

src/microcosm.js
```diff
@@ -121,23 +116,40 @@ inherit(Microcosm, Emitter, {
...
-  release (action) {
-    if (this.dirty) {
-      this.dirty = false
-      this._emit('change', this.state)
+  off (type, callback, scope) {
+    let [event, meta] = type.split(':', 2)
+
+    switch (event) {
+      case 'change':
+        this.changes.off(meta, callback, scope)
```
efatsi: need `meta || ''` here as well?

src/compare-tree/key-path.js
```diff
@@ -0,0 +1,40 @@
...
+export function getKeyStrings (array) {
+  let sorted = array.slice(0).sort(byKeyLength)
+  let key = ''
+
+  for (var i = 0, len = sorted.length; i < len; i++) {
+    key += i > 0 ? PATH_DELIMETER : ''
+    key += getKeyString(sorted[i])
+  }
+
+  return key
```
efatsi: could this whole thing be:
```js
sorted.map(getKeyString).join(KEY_DELIMETER)
```

efatsi: is the function of this just to ensure that if you call `on` with `'change:users,meta.focused'`, and then `off` with `'change:meta.focused,users'` that they would both reconcile down to the same string?

nhunzaker: Yep!

src/compare-tree/index.js
```diff
@@ -0,0 +1,139 @@
...
+      var { node, last, next } = stack.pop()
+
+      if (last !== next) {
+        node.revision += 1
+
+        var edges = node.edges
+        for (var i = 0, len = edges.length; i < len; i++) {
+          var edge = edges[i]
+
+          if (node.revision > edge.revision) {
```
efatsi: what does this check accomplish?

src/compare-tree/query.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+inherit(Query, Emitter, {
+  extract (state) {
+    let len = this.keys.length
+    let values = Array(len)
+
+    for (var i = 0; i < len; i++) {
+      values[i] = get(state, this.keys[i])
+    }
```
efatsi: i see a lot of array iteration like this in this PR. is the use of `.map` not ideal? seems like this could be written down to
```js
return this.keys.map(function (key) {
  get(state, key)
})
```

src/compare-tree/key-path.js
efatsi: i'm tempted to suggest pulling this out, makes it tougher to understand that the id for the query you generate is likely just the same thing that you've passed in to `on` to begin with. and it protects against a use case that i imagine would be quite rare, as well as not something i would really expect to be covered (turning on and off with the same key seems like a reasonable requirement).

src/compare-tree/index.js
```diff
@@ -0,0 +1,139 @@
...
+            edge.revision = node.revision
+
+            if (edge instanceof Query) {
+              edge.trigger(this.state)
+            } else {
+              stack.push({
+                node: edge,
+                last: last == null ? last : last[edge.key],
+                next: next == null ? next : next[edge.key]
```
efatsi: in what situations are these null? if a key was deleted from the state tree?

src/compare-tree/index.js
```diff
@@ -0,0 +1,139 @@
...
+              })
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
efatsi: new high score!

test/unit/compare-tree/compare-tree.test.js
```diff
@@ -0,0 +1,330 @@
...
+    it('only invokes a CompareTree once if given multiple compares', function () {
+      let a = set(SOLAR_SYSTEM, ['meta', 'selected'], 'jupiter')
+      let b = set(a, ['meta', 'focused'], false)
+
+      let handler = jest.fn()
+
+      this.tree.on(['meta.selected', 'meta.focused'], handler)
+      this.tree.update(b)
+
+      expect(handler).toHaveBeenCalledTimes(1)
```
efatsi: where are the "multiple compares" in this?

test/unit/compare-tree/compare-tree.test.js
```diff
@@ -0,0 +1,330 @@
...
+    })
+
+    it('can traverse missing state keys', function () {
+      this.tree.on('meta.selected', n => n)
+
+      this.tree.update({ meta: null })
+
+      this.tree.update({
+        meta: { selected: true }
+      })
```
efatsi: is there an assertion here that i'm not seeing?

src/compare-tree/key-path.js
efatsi: and, if you're feeling strongly about leaving it in, at minimum change it's name to reflect that it sorts the key strings by length of path depth

efatsi: but also know that this doesn't catch all cases, ie:
```js
this.tree.on(['meta.selected', 'meta.focused']
this.tree.off(['meta.focused', 'meta.selected']
```

test/unit/compare-tree/compare-tree.test.js
efatsi: ah, sorry, the assertion here is that nothing blows up, since the `meta.selected` path disappears, and then reappears

src/compare-tree/key-path.js
nhunzaker: Hmm. Cool. I'll take another pass.

src/compare-tree/index.js
```diff
@@ -0,0 +1,139 @@
...
+  getKeyStrings
+} from './key-path'
+
+import {
+  castPath
+} from '../utils'
+
+const ROOT_KEY = '~'
+
+export default function SubscriptionTree (initial) {
```
efatsi: want to call this `CompareTree`?

efatsi:
Some strong stuff in here. A couple thoughts in general though around the readability of the code:

- a lot of your methods are extremely dense with logic, particularly in CompareTree
  - in ruby land I'd recommend breaking out individual dense pieces of logic into well named private methods that self document their functionality
  - i think doing that here would bloat up that class to a pretty uncomfortable level
  - an alternative would be to break out the logic into close but separate classes (could see splitting it up by `on`/`track`/`add`, `off`/`prune`, `update`/`scan`). not sure if doing so would make keeping track of the tree easy or not
  - another alternative would be to just comment the hell out of things. looking at `prune` in particular (for which I had to read all of the tests to confirm functionality suspicions / be surprised by functionality that's covered), you're doing some very interesting things just based off what the state of various arrays are that you iterate over. looking at that code it's not obvious at all what's being accomplished - so room for improvement there.
- tough to determine which methods are in the public API and which are private, not a problem once I was fully acquainted with the code, but at the least separating those visually would be helpful

If documenting through comments is a path you want to take, let me know as that's something I can easily step in with. I could also try to take a crack at breaking out CompareTree into isolated bits of functionality if you feel good about that approach as well.

src/microcosm.js
nhunzaker: Good catch, yes.

test/unit/compare-tree/compare-tree.test.js
nhunzaker: Test name might be from a bad find/replace. This is testing to make sure that, when a subscription listens to two key paths, and both of them change at once, it only fires the callback a single time.

examples/canvas/app/boot.js
nhunzaker: Based on Slack convo, I changed this to `keys watched/frame`.

src/microcosm.js
nhunzaker: Fixed.

src/compare-tree/index.js
nhunzaker: This check verifies that I don't double emit a event if multiple paths for a subscription change. Sort of like not double requiring a dependency.

src/compare-tree/query.js
nhunzaker: It just isn't nearly as fast. It allocates memory for the anonymous function, and has to pay the penalty to maintain the JS closure.

But I could run the numbers. It might not really matter.

src/compare-tree/index.js
nhunzaker: Haha. Yeah....

src/compare-tree/index.js
nhunzaker: Fixed.

nhunzaker:
@efatsi Took another pass, but I'll resume some of the extra feedback in the morning. Thank you for taking the time to do such a thorough review.

src/compare-tree/index.js
```diff
@@ -0,0 +1,197 @@
...
+   * @param {String} id Identifier for the node.
+   * @param {String} keyPaths Each query tracks a list of key paths
+   */
+  addQuery (id, keyPaths) {
+    if (!this.nodes[id]) {
+      this.nodes[id] = new Query(id, keyPaths)
+    }
+
+    return this.nodes[id]
+  },
```
nhunzaker: I split up `add` into `addQuery` and `addNode`. I like this better, but wonder if we could just have a single node class. 

The trick is that nodes represent a single key name in a JS object, whereas Queries represent multiple pathways.

src/compare-tree/index.js
```diff
@@ -0,0 +1,197 @@
...
+          parent.edges.splice(index, 1)
+        }
+
+        if (parent.edges.length <= 0) {
+          this.prune(parent, [castPath(parent.id).slice(0, -1)])
+        }
+      }
+    }
+
+    delete this.nodes[node.id]
```
nhunzaker: This could be way shorter and easier to parse if we just sent in a `parent` property on Node and Query. Trouble is that Query can have multiple parents, but maybe it could be restructured.

Essentially I want:

```javascript
while (node) {
  if (node.parent) {
    node.parent.disconnect(node)
  }

  if (node.edges.length === 0) {
      this.nodes.delete[node.id]
  }

  node = node.parent
}
```

This function is responsible for cleaning up a tree so that it only contains the nodes for all current queries. So if we remove a query, and there are no longer any handlers, we want to "prune" the branch of key paths so that only the key paths we care about are maintained.

Thinking more on this, `Query` tracks all of its key paths, and we never remove a node outright, always as the product of removing a query. Maybe we could just traverse the key paths for the query.

test/unit/compare-tree/compare-tree.test.js
nhunzaker: That could be clearer, though. I'll see what I can do.

nhunzaker: Done.

src/compare-tree/index.js
```diff
@@ -0,0 +1,193 @@
...
+
+      do {
+        if (node.isAlone()) {
+          node.orphan()
+          delete this.nodes[node.id]
+        }
+
+        node = node.parent
+      } while (node)
+    }
```
nhunzaker: This is starting to feel better, but I don't like that there's an exception to the rules for how IDs are generated for the root node. Basically, the root node has an id of `'~'` to work around the fact that it would otherwise by an empty string (`''`), which they key path utilities convert into an empty array.

I think this exception could be eliminated. Just need to spend some more time on it.

nhunzaker: Nope... actually. We can totally do this. Change soon.

nhunzaker: done.

src/compare-tree/index.js
nhunzaker: You know what though? We only need to check this for Queries. Nodes only traverse down, and have single dependencies.

What do you think about renaming `Node` to `Key`, or some other more distinguishing name?

test/unit/compare-tree/compare-tree.test.js
nhunzaker: Updated.

nhunzaker:
Alright, I think I have all the feedback on this. @efatsi anything else you see?

efatsi:
👍 🐗

Issue #190, opened 2016-12-01
Notifications


nhunzaker:
It does not have to live on this repo, but I would like a generate abstraction for notifications. It would track a Microcosm's history object and provide detailed information to a user.

**Use cases:**

1. Flash notifications
2. a11y announcements

@ten1seven mentioned that there's probably prior art in the desktop software space. Let's leverage existing patterns.

nhunzaker:
Polled Twitter and got a few leads:

1. https://github.com/dequelabs/ngA11y#accessible-announcement-service
2. https://github.com/patrickfox/a11y_kit#announce---say-anything

nhunzaker:
Closing this out for now. This is a work in progress here:
https://github.com/vigetlabs/microcosm-notify

Maybe we can eventually move it over into this repo once it settles a bit.

PR #266, opened 2017-03-22
Pass command and status into register(). Command into intercept()

nhunzaker:
This commit updates the register method of Domains and Effects to
accept the given command and status for an action. Additionally, if a
function is returned from register, the dispatcher will use that
function as the handler.

This behavior also carries over to intercept(), however only the
command is given. This is because intercept applies _before_ an action
is created.

I primarily want this for an autosave feature on a current project:

```javascript
import debounce from 'lodash.debounce'
import {get, set} from 'microcosm'
import { save } from '../actions/api'

const DELAY = 5000

class Autosave {

  setup (repo, options) {
    let delay = get(options, 'delay', DELAY)
    this.save = debounce(repo.prepare(save), delay)
  }

  register (command, status) {
    return command.undoable ? this.save : null
  }

}
```

Related issues:
https://github.com/vigetlabs/microcosm/issues/190


docs/api/domains.md
```diff
@@ -135,5 +135,25 @@ const Planets = {
   }
 }
 
-repo.push(Actions.add, { name: 'earth' }) // this will add Earth
+repo.push(addPlanet, { name: 'earth' }) // this will add Earth
+```
+
+Registration is given the command associated with an action, and the
```
greypants: What is a "command" associated with an action?

docs/api/domains.md
```diff
@@ -135,5 +135,25 @@ const Planets = {
...
+register will be used as the handler for a Microcosm dispatch:
+
+```javascript
+import { addPlanet } from '../actions/planets'
+
+const StatusLog = {
+  //...
+  register (command, status) {
+    switch (command) {
+      case Actions.add
```
greypants: What is `Actions.add`? Where does `Actions` come from?

docs/api/domains.md
```diff
@@ -135,5 +135,25 @@ const Planets = {
...
+Registration is given the command associated with an action, and the
+current status of the action. Additionally, functions returned from
+register will be used as the handler for a Microcosm dispatch:
+
+```javascript
+import { addPlanet } from '../actions/planets'
+
+const StatusLog = {
+  //...
+  register (command, status) {
```
greypants: so register... gets called on every Domain and Effect anytime any action is pushed? Only when if it returns a function instead of an object?

docs/api/domains.md
nhunzaker: Good question. We have some terms issues. We call the functions you pass to `repo.push` an action, or at least, we store them under `./actions`, but we also call the thing returned from `repo.push` an action.

So I've started calling the things you pass to `repo.push` _commands, like:

```
repo.push(command, paramOne, paramTwo)
```

What do you think?

docs/api/effects.md
```diff
@@ -93,3 +93,27 @@ class Planets {
 repo.addEffect(Planets)
 repo.push(addPlanet, { name: 'earth' }) // this will add Earth
 ```
+
+Registration is given the command associated with an action, and the
+current status of the action. Additionally, functions returned from
```
greypants: can you link to or enumerate the possible statuses of an action here?

docs/api/effects.md
```diff
@@ -93,3 +93,27 @@ class Planets {
...
+const Notifications = {
+  //...
+  register (command, status) {
+    switch (status) {
+      case 'reject':
+        return this.alert
+    }
+  },
+  alert () {
+    alert('Danger danger!')
```
greypants: So this would run when ANY action ever is rejected?

docs/api/domains.md
nhunzaker: Would this be better as:

```
`register` is given the command passed to `repo.push`. For example, when `repo.push(addUser)` is executed, `register` will gets called with `addUser` and the current status of the action the domain is responding to.
```

docs/api/domains.md
nhunzaker: Thank you. I thought I caught all of the instances of this. I was wrong.

docs/api/effects.md
nhunzaker: Yes.

docs/api/domains.md
nhunzaker: For effects, every time. For domains, we cache the answer. So there's actually an edge case here - we should only cache on the old version of register, or not cache at all. Otherwise, I think the expectation would be that this gets called every single time, when it wouldn't for Domains (because of caching).

Hmmm. This feature is really for effects, not Domains. What do you think about only doing this for Effects? Is the inconsistency there confusing?

docs/api/domains.md
greypants: hmm... I think that makes some sense, but it that change has pretty wide documentation and api implications. `intent` was _just_ renamed to `action`, which seems like now it'd need to be changed again to `command`.

I agree it needs some distinction though, and don't really have a better idea.

nhunzaker: Haha... Yeah. I know... I'm not sure where to go from here. Just calling this `action` instead of `command` works really well for me. It would be better to rename the `Action` class, which doesn't have a public API.

docs/api/domains.md
greypants: I think it'd be fine to make this just for Effects, until we discover a pressing need to extend it to Domains. You can always push an ~~action~~ command from an effect to update state. It definitely makes more sense in my head for Effects.

docs/api/domains.md
nhunzaker: I've filed an issue to talk about sorting out this nomenclature:
https://github.com/vigetlabs/microcosm/issues/268

The gist of it is - I think I'm wrong. We should just call actions _actions_, not commands. The thing we get back from `repo.push()` should be called a separate thing. That prevents a breaking change and there's no reason to change existing code.

nhunzaker:
I'm going to close this out for now and revisit it. There's a few things I actually still need to figure out.

PR #270, opened 2017-03-24
Typo in Presenter docs

efatsi:


nhunzaker:
Why is this a typo?

docs/api/presenter.md
```diff
@@ -41,7 +41,7 @@ class PlanetsPresenter extends Presenter {
 
   getModel (props, state) {
     return {
-      planets: data => data.planets
+      planets: state => state.planets
```
nhunzaker: Sorry... responded too soon. We have a pickle here. getModel accepts presenter props and state as arguments, however we also have the first argument of model bindings, which is repo state.

Everywhere else, we're probably using `state => state.planets`. What are your thoughts on how we can make this less confusing?

efatsi: Ah, yeah sorry myself, this change is definitely not ideal. We can probably leave this as is for now.

Issue #268, opened 2017-03-24
Action nomenclature


nhunzaker:
The term `action` is currently overloaded. I have a bad habit of changing names on this project, something I don't want to do too much more of in the future, but the term `action` is really overloaded and I get a lot of questions clarifying what is an action vs whatever you get from `repo.push(thing)`. 

I don't want a breaking change, but want to make the distinction clearer between these two pieces:

```javascript
// 1 - What do we call this?
function addUser (params) {
  return fetch.post('/users', params).then(response => response.json())
}

// 2 - What do we call this?
let whatToCallThis = repo.push(addUser, { name: 'Billy' })
```

Here's what I'm thinking:

1. The thing you pass to `send()` and `repo.push()` is an action. There's no change here.
2. The thing you get back from `repo.push()` is a _Task_

I think this sounds very natural, but we might want to add some aliases for a few of the methods that update a task. The sequence could be:

- `task.start()` - aliases `task.open()`
- `task.update()` - no change
- `task.done()` - aliases `task.resolve`
- `task.error()` - aliases `task.reject`
- `task.cancel` - no change

We already have these aliases for actions, so you can subscribe to `action.done` and `action.resolve`. This also increases consistency with `task.onDone` and `task.onError`, which might be a good change to make regardless.

Overall, this could look like:

```javascript
function wait (time) {
  return task => {
    task.start()

    let timeout = setTimeout(task.done, time)

    task.onCancel(() => clearTimeout(timeout))
  }
}

let task = repo.push(wait, 1000)

// Just kidding!
task.cancel()
```

mackermedia:
👍

h0tl33t:
I like `task` for this a lot!  🥇

keturn:
If "the thing you get back from repo.push" is the single asynchronous result of some one-time action, and any callbacks you register on it will be called no more than once (unlike the callbacks you register from Domains), that sounds a lot like a Promise (or Deferred).

Not all implementations of Promise have _abort_ and _update_ methods (the version of Promise that made it in to ES6 has neither, and Bluebird seems to have decided that `cancel` should mean "ignore results" rather than "abort operation in progress"), but there's certainly precedent for both.

nhunzaker:
> Bluebird seems to have decided that cancel should mean "ignore results" rather than "abort operation in progress")

[...]

PR #271, opened 2017-03-27
Rename Actions to Tasks, replace Prototypes with Classes

nhunzaker:
Oof. I did it. This commit renames Actions to Tasks. It also does a few other things
while I was at it:

### Replace Prototypes with Classes

Originally I avoided using classes because of Bloat bloat, however using Buble for transpilation appears to have no effect on the end gzipped size.

### Glossary document

I've also added a glossary of terms for Microcosm. This should help us keep things straight, and aid us in identifying areas where we can decrease the amount of Microcosm specific stuff you need to know.

### Actions aren't going away

Just to clarify, the idea of actions isn't gone. I just want a stronger distinction between what you pass to `repo.push()` and what you get as a result. Eventually, I want to be able to have Tasks manage multiple actions. A Task is composed of actions, and tracks their progress.

---

Related issues: https://github.com/vigetlabs/microcosm/issues/268


nhunzaker:
I'm going to close this out and send out some more focused PRs. There's no way I'll ask anyone to review 100 files.

Issue #229, opened 2017-02-23
action.parallel


nhunzaker:
What would it look like to batch actions together in parallel?

```
action.parallel(
  initial => first(initial),
  next  => second(next)
).onDone(doSomethingElse)

// or
action.parallel(first, second).onDone(doSomething)
```

Should we just lean on Promise.all, or a similar idea?

nhunzaker:
Subsquently, we could follow up with a `repo.waterfall` and `repo.parallel`. Need to figure out how that fits in to everything else naming wise.

keturn:
Another place where it seems like this fits in neatly to a Promise interface. Your `.parallel` much like [`.all`](http://bluebirdjs.com/docs/api/promise.all.html) or [`.props`](http://bluebirdjs.com/docs/api/promise.props.html) (though Angular's `$q.all` takes either the Array or key-value Object signature).

`.waterfall` is not something I've seen show up in a Promise interface so much, although `q` does [talk about that pattern](http://documentup.com/kriskowal/q/#sequences) and boils it down to `[first, second].reduce(Q.when)`

nhunzaker:
Updating this issue now that we have action sequences using generators in `12.6.0-beta`:

```javascript
export function createMessageAndGoToEditor () {
  return function * (send) {
    let response = yield send(create)

    let message = get(response, 'data.message')

    yield send(setMessageProperties, message)

    yield send(save)

    yield send(navigate, `/messages/${message.id}`)

    yield send(setStep, 'editing')
  }
}
```

I'm pretty happy with this. These sequences can be composed together very easily, and you get an action back that represents the entire workflow. Long term, I could see us shipping a set of addons for simple control flow. Something like:

```
function confirmAndDelete (user) {
  return function * (send) {
    yield send(confirm, `Are you sure you want to delete ${user.name}?`)
    yield send(deleteUser, user.id)
  }
}
```

@keturn I like your idea of just making, what this issue calls `waterfall`, work like `Promise.all`. One thought I had for the future was to always pass the `repo` as the first argument of an action creator. If you returned an action from that, we'd just use that action instead of creating one under the hood:

```javascript
function fetchUser (repo, id) {
  return fetch(`/users/${id}`).then(res => res.json())
}

function fetchUsers (repo, listOfUsers) {
  let requests = listOfUsers.map(id => repo.push(fetchUser, id))

  return Promise.all(requests)
}
```

[...]

PR #272, opened 2017-03-27
Rename action to task

nhunzaker:
This commit renames Actions to Tasks. Additionally:

**Glossary document**

I've also added a glossary of terms for Microcosm. This should help us
keep things straight, and aid us in identifying areas where we can
decrease the amount of Microcosm specific stuff you need to know.

**Actions aren't going away**

Just to clarify, the idea of actions isn't gone. I just want a
stronger distinction between what you pass to repo.push() and what you
get as a result. Eventually, I want to be able to have Tasks manage
multiple actions. A Task is composed of actions, and tracks their
progress.


docs/api/glossary.md
```diff
@@ -0,0 +1,39 @@
...
+app.
+
+**Task:** A unit of work that needs to be performed. Tasks execute the
+action they are created with. Based on the outcome of the action, they
+can move into an `open`, `update`, `resolve`, `reject`, and `cancel`
+state. As a task's state changes, a Microcosm uses Domains to
+determine how it should turn the result of the Task's work into new
+repo state. Effects can respond to changes in task states to
+perform side-effects, like saving to localStorage or tracking user
+behavior using analytics tool.
```
nhunzaker: @leobauza @greypants @efatsi Feels? We could blow this out further.

efatsi:
> Eventually, I want to be able to have Tasks manage multiple actions. A Task is composed of actions, and tracks their progress.

Can you elaborate on that?

greypants:
Are we still calling what you `push` an action though?

nhunzaker:
@greypants Yes. An action is what you send into `push`, no need to rename the `actions` folder :). 

@efatsi I might be ahead of myself. See https://github.com/vigetlabs/microcosm/issues/229. Basically, I want away to create a Task to be able to manage multiple sub-tasks, triggering `onDone`, `onError`, etc. based on the result of the whole.

nhunzaker:
@efatsi For example, a confirmation warning for deleting a record. What if we could abstract that into a `confirm` action and a `deleteRecord` action?

```javascript
function confirmAndDeleteRecord (id) {
  return (task, repo) => {
    task.sequence([
      () => repo.push(confirm, 'Are you sure'), 
      () => repo.push(deleteRecord, id)
    ])
  }
}
```

I don't have ergonomics figured out here, but that's sort of where my head is at.

docs/api/glossary.md
leobauza: I like this, we should definitely have this. Could we move it under the `guides`/`introduction` section? Then we could break up the glossary into the different sections like `core api`, `addons`, `general stuff?`, etc

nhunzaker: Cool. I like that a lot.

efatsi:
I might be alone here, but this change confuses me. So you still `push` an `action`, but then the `action` turns into a `task`? Or a `task` becomes in charge of the `action`?

I'd like to further understand the benefit from making this nomenclature change (functionally things are staying the same right?) because it definitely brings with it the cognitive load of understanding two components instead of one for the "this is how things change" functionality.

nhunzaker:
@efatsi I was thinking a task becomes in charge of the action. I'll hold off merging this. Let's get this right.

nhunzaker:
Going to hold off on this for now.

Issue #254, opened 2017-03-10
Domain and Effect creation


nhunzaker:
> *Leo*: Why is there no code sample for addDomain like there is for addEffect ?

> *Nate*: There is no reason. I'll add an issue for this. We also need to better describe how you can create domains and effects (simple objects vs classes)

keturn:
👍 for highlighting "simple objects vs classes," I tripped over that one for a while.

nhunzaker:
@keturn I've sent out a PR that documents the creation process here: 
https://github.com/vigetlabs/microcosm/pull/275

I'm happy to help clear things up. Would you be willing to share what tripped you up, so I can make sure to cover it in the docs?

keturn:
The way I remember it is that I thought "both Domain and Effect have `register` methods that refer to methods on `this.`, I'll make them both classes",

and then things confusingly Did Not Work until I made the Domain a simple object instead. 

However, the documentation in this patch says you can use either approach with either thing. Is that a change since version 11?

nhunzaker:
Hmmm. It looks like we added it in 10.2.0:
https://github.com/vigetlabs/microcosm/blob/master/CHANGELOG.md#1020

So I'm not sure. We use the `class` form on a project that's still on 11.x without a problem.

Either way, I am glad that the class form jives with how you were thinking about Domains and Effects!

nhunzaker:
Just merged a related PR. I'm going to close this out.

PR #275, opened 2017-03-28
Add documentation on creating effects and domains

nhunzaker:
Also:

- Use new nested syntax for domain status registration
- Add similar entry for domain creation in microcosm.md

Related issues:

- https://github.com/vigetlabs/microcosm/issues/254

---

Documentation

No

- [x] All tests are passing

### DIFF
--- a/docs/api/domains.md
+++ b/docs/api/domains.md
@@ -1,8 +1,9 @@
 # Domains
 
 1. [Overview](#overview)
-2. [Subscribing to different action states](#subscribing-to-different-action-states)
-3. [API](#api)
+2. [Creating Domains](#creating-domains)
+3. [Subscribing to different action states](#subscribing-to-different-action-states)
+4. [API](#api)
 
 ## Overview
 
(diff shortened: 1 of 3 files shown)


docs/api/domains.md
```diff
@@ -23,6 +24,67 @@ Domains do not enforce any particular structure. However specific
...
+  },
+  handleAction (state, payload) {
+    let newState = { ...state, prop: payload.prop }
+
+    return newState
+  },
+  register () {
+    return {
+      [action] : this.handleAction
+    }
```
leobauza: There should be a mention that the registering can happen like this or how it is shown later (splitting it into it's lifecycle keys). Since this is already a confusing part seeing it written differently all over might be confusing.

leobauza:
Nested syntax for domain status registration is less confusing and verbose, I like that. 

In seeing all this we need to combine some of what is in the `actions` section and in the `domains` section of the docs to explain registering in a single consistent way. 

Maybe it's just a matter of changing how it's written everywhere and explaining `register()` separately. I think adding a note that explains that `[action]: this.handleAction` is the same as `[action.done]: this handleAction` which is the same as `[action]: {done: this.handleAction}` (assuming I am getting that right) would solve this problem for this PR until we rework the sections.

nhunzaker:
@leobauza you're totally getting that right. I wonder what the best way to handle that would be.

leobauza:
@nhunzaker well, this is a microcosm pattern right? So under `architecture` under `Domain` and `Effects` we should add a new section called `The Register Method`. With a full explanation of what I said on my other comment. Then you can write the `register()` method in any of it's three forms as long as we add a note that says something like:

"Confused by the `register()` method? the architecture section should clear things up" 

And we link to that from there. We can also do more explaining about it's origin and how it compares to Redux's switch statements on reducers.

nhunzaker:
Awesome. Sounds good!

Issue #232, opened 2017-02-24
Formal API for all repo history done


mackermedia:
Recently I ran into an issue with a test that was trying to assert before my nested chain of action promises had completed.

I solved it with: `Promise.all(this.repo.history.toArray()).then(() => { ...assertion }`.

It might be nice to have a helper function on `repo` to handle this.

nhunzaker:
Shipped.

PR #276, opened 2017-03-28
Add history.wait() feature to wait for all outstanding actions

nhunzaker:
On one of our current projects, we have a bit of a hack to wait for all actions to complete before running a test. I want to more formally support that, and add some documentation around how it works. 

This commit adds promise interop for action history. This is useful for waiting for history to complete before doing some other type of work. It also adds documentation for history, which is somewhat unstable still, something we clarify in the doc.

With this PR, you can now do:

```javascript
describe('An AJAX behavior', function() {

  it('adds a user', async function () {
    let repo = new MyMicrocosm({ maxHistory: Infinity })

    repo.push(getUser, 1)
    repo.push(getUser, 2)
    repo.push(getUser, 3)

    await repo.history

    expect(repo.users.length).toEqual(3)
  })

})
```

I need way more infographics for this. Please help!

---

Related issues:
https://github.com/vigetlabs/microcosm/issues/232

### DIFF
--- a/docs/api/history.md
+++ b/docs/api/history.md
@@ -3,9 +3,10 @@
 1. [Overview](#overview)
 2. [API](#api)
 
-**Note: This is a work in progress document and feature. History has been a
-private API for a very long time. We're still working through the best
-public interface for working with Microcosm's action history.
+**Note: This is a work in progress document and feature. History has
+been a private API for a very long time. We're still working through
+the best public interface for working with Microcosm's action
+history.**
 
 ## Overview
 
--- a/docs/api/history.md
+++ b/docs/api/history.md
@@ -106,6 +106,15 @@ In the example above, actions `one` and `two` will be forgotten. The
 repo's history will reconcile, and it will be as if they never
 existed.
 
+This flips the `disabled` state of each action provided. By executing
+toggle a second time, these actions will be re-enabled:
+
+```
+// Actions disabled in the prior example
+repo.history.toggle([ one, two ])
+// `one` and `two` have been re-enabled
+```
+
 ### toArray()
 
 Return an array version of the active branch:
--- a/docs/api/history.md
+++ b/docs/api/history.md
@@ -1,0 +1,182 @@
+# History
+
+1. [Overview](#overview)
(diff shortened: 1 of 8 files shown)


docs/api/history.md
```diff
@@ -0,0 +1,182 @@
...
+let three = repo.push(action)
+```
+
+The head of the history tree above is now three. Microcosm will
+calculate state by reconciling `root`, `one,` and `three`.
+
+
+### toggle([ ...actions ])
+
+Disable a group of actions:
```
efatsi: does `toggle` work both ways? if not should this be something like `delete`?

nhunzaker: Toggle works both ways. Thanks for pointing that out. I've also considered adding a `forget` and `remember` method, so you can do this more deterministically. Thoughts?

efatsi: ah nice, i'd recommend documenting that it goes both ways. I have no idea what the use case for this would be so no opinion on the more targeted methods

nhunzaker: We do it on a certain project for undo/redo:

undo redo-example

efatsi:
Code looks great to me 👍

test/unit/history/then.test.js
```diff
@@ -0,0 +1,40 @@
...
+    let two = history.append(action)
+    let three = history.append(action)
+
+    setTimeout(function() {
+      one.resolve()
+      two.resolve()
+      three.resolve()
+    }, 10)
+
+    return history
```
mackermedia: no expectation here?

mackermedia:
Kinda weird to see tests without explicit assertions, maybe that's an okay thing. Will an unhandled promise cause the test suite to fail?

Otherwise, 👍

test/unit/history/then.test.js
nhunzaker: Ah this could be clearer. I'm expecting that history resolves correctly. I should add a comment.

nhunzaker:
@mackermedia A rejected promise will fail a Jest test if you return it from the test function.

PR #278, opened 2017-03-29
Add dedicated domain class

nhunzaker:
I don't want to merge this yet, but I thought I could at least get some eyes on it.

This is the formal class for domains. Eventually, I plan to add an API for adding sub-domains. This is just the bridge work.


greypants:
Seems reasonable. In the future would you `class FooDomain extends Domain`, or just create a plain `class` like you do with effects?  Either way, it'd be nice to be able to create Domains as classes, if for no other reason than consistency with Effects.

nhunzaker:
I was going to support both use cases initially. The benefit of extending from Domain would be an additional set of library features like `addSubDomain`. I'm even debating allowing parent domains to dispatch to children.

src/domain.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * Domains define the rules in which resolved actions are converted into
+ * new state. They are added to a Microcosm instance using `addDomain`.
+ * @constructor
+ * @param {Microcosm} repo The instance of Microcosm that added this domain.
+ * @param {Object} [options] Options passed from `addDomain`
+ */
+export default function Domain (options, repo) {
+  this.options = options
+  this.repo = repo
```
efatsi: my js-fu isn't strong enough to understand how defining this function does anything, but does the fact that the order is `options, repo` here and `repo, options` in `setup` have any significance?

nhunzaker:
I'm going to revise and resubmit this.

PR #279, opened 2017-03-30
General action refactor

nhunzaker:
Changes related to actions:

- Action methods are assigned manually, instead of looping over a map
  of statuses
- Renamed `action.disposable` to `action.complete`
- History listens to action changes, instead of actions telling
  history to reconcile

Other changes:

- The uglify step now mangles propertes that start with
  underscores, like: `_callOrSubscribeOnce`

Interestingly, there is 100 more lines of code, but the gzipped payload is actually smaller.


src/action.js
```diff
@@ -1,68 +1,187 @@
...
 
-import {
-  STATES
-} from './constants'
+let uid = 0
 
 /**
- * Actions encapsulate the process of resolving an action creator. Create an
- * action using `Microcosm::push`:
+ * Actions represent the state of an action. As actions change, their
```
mackermedia: `Tasks represent the state of an action`?

nhunzaker: Haven't merged that one yet :-/

src/get-registration.js
```diff
@@ -1,5 +1,17 @@
...
+  inactive  : 'inactive',
+  open      : 'open',
+  update    : 'loading',
+  loading   : 'update',
+  done      : 'resolve',
+  resolve   : 'done',
+  reject    : 'error',
+  error     : 'reject',
+  cancel    : 'cancelled',
+  cancelled : 'cancel'
```
mackermedia: are a bunch of these intentionally flip-flopped?

mackermedia:
Hard to understand the context around this change. Code looks 👍  except for one question on `STATUSES`.

src/action.js
mackermedia: As-is it reads kind of odd: `Actions represent the state of an action.`

nhunzaker: That's why I want to rename it :P, but I can revert this comment back.

src/get-registration.js
nhunzaker: Ah. Why did I do this.  This should still be called aliases.

This is so that if you subscribe to `reject`, or `error`, they hit the same handler.

nhunzaker:
@mackermedia I want to eventually move over to using es6 classes. Assigning directly to the prototype is pretty wonky looking for that case. Also, I didn't like how a bunch of methods were getting added to Action from a configuration file. I thought it would make it easier, but it just made it harder to understand.

mackermedia:
I wouldn't say it made it harder to understand... I don't know enough about the general codebase to say that. Generally I'd prefer to work with es6 classes over prototypes, so you have my 👍

src/action.js
```diff
@@ -1,68 +1,187 @@
...
+
+  /**
+   * Subscribe to when an action opens.
+   * @param {Function} callback
+   * @param {*} [scope]
+   */
+  onOpen (callback, scope) {
+    this._callOrSubscribeOnce('open', callback, scope)
+    return this
+  },
```
efatsi: is there existing code somewhere that invokes this?

nhunzaker: ActionButton and ActionForm:
https://github.com/vigetlabs/microcosm/blob/master/src/addons/action-button.js#L25
https://github.com/vigetlabs/microcosm/blob/master/src/addons/action-form.js#L53

Issue #280, opened 2017-03-31
Strict Mode


nhunzaker:
Microcosm has quite a few assertions under the hood to catch for impossible states. For example, this assertion runs to make sure the `register()` method is working with a valid action status:

```
console.assert(STATUSES[status], 'Invalid action status ' + status)
```

I'd like to expose these in a `microcosm.strict.js` build. Presently, they get stripped out using `babel-plugin-unassert`. As we discover undesirable states, we could continue to add such assertions. We might even go so far as to add warnings, which won't raise an exception, just help with using Microcosm the way we intend it to be used.

I see a few tasks:

- [ ] Create a `microcosm.strict.js` build that does not remove assertions. 
- [ ] Add an `installation.md` guide to our docs, with information on strict mode, how to use it, and how to remove assertions in production. This could just be by providing an example Webpack module alias.

We also need to be really intentional about reporting unexpected behavior, so we can add support in this strict build.

This is somewhat related to https://github.com/vigetlabs/microcosm/issues/250

mackermedia:
Would strict-mode by default be the most helpful to new/unknowing microcosm devs?

nhunzaker:
I think so. Yes. I'm just hesitant to end up with assertions running in production that check for things that aren't necessarily an error, but poor usage.

nhunzaker:
Sorry... what's the best way for us to accomplish this?

PR #285, opened 2017-04-04
Add strict mode

nhunzaker:
This commit adds strict mode: a special bundle of Microcosm that
includes development-only assertions. It also adds an installation
guide for setting up strict mode with Webpack.

Fixes:
https://github.com/vigetlabs/microcosm/issues/280


nhunzaker:
@leobauza We probably want to flesh out the `installation.md` doc more. This is just the bare minimum I thought necessary.

PR #286, opened 2017-04-05
Generators define sequences of actions

nhunzaker:
Often times we need to dispatch multiple actions in sequential order. For example, what if we want to ask the user to confirm their action before deleting a record?

This can be accomplished by using a generator:

```javascript
function ask (message) {
  return action => {
    if (confirm(message)) {
      action.resolve()
    } else {
      action.reject()
    }
  }
}

function deleteUser (id) {
  return fetch.delete('/users/${id}').then(response => response.json())
}

function confirmAndDelete (user) {
  return function * (send) {
    yield send(ask, `Are you sure you want to delete ${user.name}?`)
    yield send(deleteUser, user.id)
  }
}
```

Each `yield` in the generator processes sequentially. A parent action is returned from `repo.push()` to represent the entire sequence. If any action is cancelled or rejected along the way, the parent action is rejected or cancelled with the same payload.

When all steps of the generator complete, the payload of the parent action will be the resolved payload of the final action.

### DIFF
--- a/docs/api/actions.md
+++ b/docs/api/actions.md
@@ -47,7 +47,7 @@ on how you author action creators.
 
 ## Writing Action Creators
 
-There are three ways to write action creators in Microcosm, all of
+There are four ways to write action creators in Microcosm, all of
 which relate to the value returned from functions passed into `repo.push()`.
 
 ### Return a primitive value
@@ -103,6 +103,45 @@ function readPlanets () {
 
 repo.push(readPlanets)
 ```
+
+### Return a generator
+
+Often times we need to dispatch multiple actions in sequential
+order. For example, what if we want to ask the user to confirm their
+action before deleting a record?
+
+This can be accomplished by using a generator:
+
+```javascript
+function ask (message) {
+  return action => {
+    if (confirm(message)) {
+      action.resolve()
+    } else {
+      action.reject()
+    }
+  }
+}
+
+function deleteUser (id) {
+  return fetch.delete('/users/${id}').then(response => response.json())
+}
+
+function confirmAndDelete (user) {
+  return function * (send) {
+    yield send(ask, `Are you sure you want to delete ${user.name}?`)
+    yield send(deleteUser, user.id)
+  }
+}
+```
+
+Each `yield` in the generator processes sequentially. A parent action
+is returned from `repo.push()` to represent the entire sequence. If
+any action is cancelled or rejected along the way, the parent action
+is rejected or cancelled with the same payload.
+
+When all steps of the generator complete, the payload of the parent
+action will be the resolved payload of the final action.
 
 ## Dispatching to Domains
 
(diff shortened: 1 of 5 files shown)


docs/api/actions.md
```diff
@@ -104,6 +104,45 @@ function readPlanets () {
...
+    }
+  }
+}
+
+function deleteUser (id) {
+  return fetch.delete('/users/${id}').then(response => response.json())
+}
+
+function confirmAndDelete (user) {
+  return function * (send) {
```
cwmanning: > In order to use Generators you must include the Babel polyfill.

does Microcosm take care of this? curious if supporting it require a certain Babel configuration/polyfill in each app

nhunzaker: Ah yes. This will require regenerator. We can't take care of this, nor I think we should.

cwmanning: is there already a section in the docs about "dependencies" like this? might want to add a note about support for these here

nhunzaker: Ah. Good call.

nhunzaker: How's this feel?
https://github.com/vigetlabs/microcosm/blob/master/docs/api/actions.md#return-a-generator

https://github.com/vigetlabs/microcosm/blob/master/docs/guides/installation.md#extra-dependencies

cwmanning: Looks good!

efatsi: I wonder if it would be a little less cryptic to have this function accept a `repo` instance, and the yield calls would be:
```js
yield repo.push(ask, `blabalbla`)
yield repo.push(deleteUser, user.id)
```

src/coroutine.js
```diff
@@ -25,6 +26,36 @@ export default function coroutine (action, body, repo) {
...
+    }
+
+    function progress (value) {
+      if (value instanceof Action) {
+        value.onDone(step)
+        value.onCancel(action.cancel, action)
+        value.onError(action.reject, action)
+      } else {
+        step(value)
+      }
```
efatsi: I think asserting that an action has been passed in would be nice here, you could make the functionality of this a little more descriptive as well:
```js
function progress (subAction) {
  subAction.onDone(step)
  subAction.onCancel(action.cancel, action)
  subAction.onError(action.reject, action)
}

File: src/coroutine.js [hunk 3]
```diff
@@ -25,6 +26,36 @@ export default function coroutine (action, body, repo) {
     return action
   }
 
+  if (body && get(body, 'constructor.name') === 'GeneratorFunction') {
+    action.open()
+
+    let iterator = body(repo.push.bind(repo), repo)
```
efatsi: what's the second `repo` argument here for?

nhunzaker: So you can reference the repo. I think just passing repo in here makes a ton of sense.

nhunzaker:
@efatsi I'll send out another PR with the "just pass repo" change. I'm not sure why I didn't think of that before.

src/coroutine.js
nhunzaker: Let's just drop this for `body(repo)` It's way clearer.

PR #289, opened 2017-04-06
Rework generator sequence

nhunzaker:
cc @efatsi 

It doesn't make sense to obscure the repo during an action generator sequence. Let's just pass it straight in:

```javascript
    function sleep (time) {
      return action => {
        setTimeout(() => action.resolve(true), time)
      }
    }

    function dream (time) {
      return function * (repo) {
        yield repo.push(sleep, time)
        yield repo.push(sleep, time)
        yield repo.push(sleep, time)
      }
    }

    repo.push(dream, 100)
```

### DIFF
--- a/test/unit/middleware/generator-middleware.test.js
+++ b/test/unit/middleware/generator-middleware.test.js
@@ -23,22 +23,20 @@ describe('Generator Middleware', function () {
     })
   })
 
-  it('continues if yielding a non-action value', function () {
+  it('throws if yielding a non-action value', function () {
     expect.assertions(1)
 
     let repo = new Microcosm()
     let step = n => n
 
     function test () {
-      return function * (send) {
+      return function * (repo) {
         yield repo.push(step, 1)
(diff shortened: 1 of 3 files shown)


test/unit/middleware/generator-middleware.test.js
```diff
@@ -187,4 +185,37 @@ describe('Generator Middleware', function () {
     await repo.history.wait()
   })
 
+  it('can push new actions during history.wait() while processing a generator', async function () {
```
efatsi: seems like a bit of an obscure test, would things blow up if some thing were to go awry in microcosm?

nhunzaker: Eh, they shouldn't. Actually, I wrote this test while I was tracking down a bug on a project. Totally unrelated. This was purely for my own confirmation.

We could remove this.

nhunzaker:
@efatsi Thanks man.

PR #293, opened 2017-04-11
Replace history.invoke with event emissions

nhunzaker:
Instead of having the history track a list of microcosms, history now
emits events:

- append: when an action is created
- update: when an action has changed
- remove: when an action has archived or been removed
- reconcile: when a history has completed reconciliation
- release: when a history has done all work, and it is safe to emit a change event

The release and reconcile events used to be the same operation, however by keeping them separate, we can start to expose the ability to control the update process. Specifically, I want to support a future feature where users can implement batching.


mackermedia:
👍

nhunzaker:
Thanks, gents!

PR #295, opened 2017-04-12
Add batching

nhunzaker:
This commit exposes control over the release process so that changes can be batched together. It also ensures that this does not break `history.wait()`.

Additionally, it updates the canvas example to use a requestIdleCallback batching strategy. This improves write throughput about 60-80%, but I'm mostly looking forward to preventing duplicate renders for loading states that finish really quickly.

http://probable-pocket.surge.sh/


examples/canvas/app/updater.js
```diff
@@ -0,0 +1,28 @@
...
+
+  return update => {
+    if (frame == null) {
+      frame = requestIdleCallback(() => {
+        frame = null
+        update()
+      }, options)
+    }
+  }
+}
```
nhunzaker: It would be cool for this to be a `microcosm-batch-updates` module, or something.

docs/api/microcosm.md
```diff
@@ -5,13 +5,58 @@
 
 ## Overview
 
-Microcosm is [Flux](https://facebook.github.io/flux/) with first-class
-actions and state sandboxing.
+The Microcosm class provides a centralized place to store application
+state, dispatch actions, and track changes.
```
cwmanning: 👍

docs/api/microcosm.md
```diff
@@ -5,13 +5,58 @@
...
+   how Microcosm releases change events. This is useful for batching
+   together frequent state changes. See
+   the [Batch Updates](../recipes/batch-updates.md) recipe for more
+   information.
+
+Feel free to add additional options to fit your use case. Any options
+you provide to Microcosm are passed into the `setup` lifecycle method:
+
+```javascript
+import Autosave from './effects/autosave'
```
cwmanning: can this example code demonstrate the concept without introducing effects? to keep the section as simple as possible

docs/recipes/batch-updates.md
```diff
@@ -0,0 +1,132 @@
+# Batch Updates
```
cwmanning: this doc is really good, 👏

src/history.js
```diff
@@ -2,25 +2,39 @@ import Action from './action'
...
+}
+
 /**
  * The central tree data structure that is used to calculate state for
  * a Microcosm. Each node in the tree is an action. Branches are
  * changes over time.
  * @constructor
  */
-export default function History (limit) {
+export default function History (config) {
```
cwmanning: with this change, would it make sense to warn when config is not an object?

docs/api/microcosm.md
nhunzaker: Hmm. Good idea.

nhunzaker:
@cwmanning good call on the test. I discovered a bug where you could get stuck in `wait()` if you waited _during_ a release.

Which made me realize... we don't need to request any more "updates" if the last update hasn't fired yet. That can reduce the complexity of the updater significantly.

Instead of doing:

```
// A lot of browsers don't support requestIdleCallback, so we patch it
import 'ric'

// Never let the user wait more than 24 milliseconds for an update
const options = { timeout: 24 }

export default function requestIdleBatch () {

  // Batching strategies return a function. This allows you to
  // maintain state within the closure above. Here, we keep track of
  // the last frame of work
  let frame = null

  return update => {
    if (frame == null) {
      frame = requestIdleCallback(() => {
        frame = null
        update()
      }, options)
    }
  }
}
```

We could just do:

// A lot of browsers don't support requestIdleCallback, so we patch it
import 'ric'

// Never let the user wait more than 24 milliseconds for an update
const options = { timeout: 24 }

```
export default function requestIdleBatch () {
  return update => requestIdleCallback(update, options)
}
```

I still want to return a function, because it would allow you to do things like setup logging or some other system. But you don't have to keep track of prior frames, which is really nice.

What do you think?

nhunzaker:
@cwmanning technically now, you could also just do:

```javascript
export default function requestIdleBatch () {
  return update => setTimeout(update, 24)
}
```

cwmanning:
@nhunzaker I like the simplified updater example. Might include the really simple one (`setTimeout`) in the example code as well, for those who may opt to avoid a polyfill.

src/history.js
nhunzaker: I don't _think_ so. History is never instantiated directly, it always passes through new Microcosm()

cwmanning: oh I see, sounds good. somehow I thought I was looking at the Microcosm constructor

nhunzaker:
@cwmanning Done, what do you think?

nhunzaker:
@leobauza Do these documentation updates feel good? Anything we should change?

docs/recipes/batch-updates.md
```diff
@@ -0,0 +1,119 @@
...
+problematic for testing. For example:
+
+```javascript
+import App from 'src/views/application'
+import { mount } from 'enzyme'
+
+test("it increases the number when the stepper is clicked", function () {
+  let app = mount(<App />)
+
+  app.find('#stepper).simulate('click')
```
efatsi: should be `app.find('#stepper')` (missing closing quote mark after `#stepper`)

efatsi:
Neat stuff, curious what peoples' thoughts are around simplifying the interface for this, and just exposing a configurable batch delay option that would instantiate an in-house updater. Reading through the Batch Updates docs is neat, but by the end of the docs I got the sense that there's really an optimal way to do things (use `requestIdleCallback`). Worth flushing this out a little more and replacing the idea of defining your own updater with just defining a batch interval? Or are there other benefits to defining your own updater that I'm not picking up on.

docs/recipes/batch-updates.md
nhunzaker: Thank you!

mackermedia:
:+1: Those docs are 💯 . Also, I think @efatsi is on to something with providing a simple mechanism to use the default recommended batch update operation.

nhunzaker:
@mackermedia @efatsi Hmm, what do you think about making this option `batch: true`, and if it's a function, use that function instead?

nhunzaker:
`batch:true` with the caveat that you'd need to polyfil requestIdleCallback

mackermedia:
i was thinking `batch: true` would do the `ric` method and you wouldn't have to think about polyfilling (microcosm happily does that for you).
or maybe also if `batch: 12` is a number it'll do the `setTimeout` method? is there any value to that method and if so, it might be nice for it to be a simple config option.
or you could also provide your own function?
overboard?

efatsi:
agreeing with @mackermedia, passing in functions allows for greater control but requires more underlying knowledge of microcosm, mainly - what's passed in to that function. My thinking would be something like `batchInterval: 12`.

docs/api/microcosm.md
leobauza: I am wondering if we should keep the reference to Flux for the "This is like X" aspect of documentation, otherwise this is an awesome change. 👍

nhunzaker:
@efatsi, @mackermedia Here's my thinking. I want a safe default, with an escape hatch. I think we can do this without exposing a lot of complexity.

I think we need to add 2-3 options:

```
batch: true
batchInterval: 24, // ? Maybe... is this useful?
updater: buildInUpdater
```

The updater function will get the options passed into Microcosm. Our default updater function would look like:

```javascript
function requestIdleBatch (options) {
  if (options.batch) {
    return update => requestIdleCallback(update, { timeout: options.batchInterval })
  } else {
    return update => update()
  }
}
```

This allows you:

1. To just set `batch: true`. End of story
2. To configure the batch interval (is it an interval, or a timeout?)
3. To configure a custom updater.
4. To conditionally apply that batch updater based on environment

I'm still debating whether or not we need `batchInterval`. I think a timeout of `24` would be exactly what I would do on any given project, and tweaking that timeout under the hood wouldn't be a breaking change (unless we go from like 24ms to 200ms).

What do you think?

mackermedia:
Honestly, I don't really know of the use-case of when I'd reach for it. It sounds like when the app starts to get janky from a lot of actions resolving and re-rendering when I want an animation to finish. 
I think the `batch: true` and `updater: myUpdateFunction` are probably all that's necessary for now?
Keep it simple?

nhunzaker:
@mackermedia Yep. That's precisely the use case. When you have a lot of actions firing, causing meaningful changes, and the successive change events cause expensive repaints.

Just having `batch` and `updater` sound good to me.

efatsi:
👍  to `batch` and `updater`, and then to this PR

nhunzaker:
`batch` and `updater` it is.

Issue #144, opened 2016-09-16
Autobind action operations


nhunzaker:
It would be nice to auto-bind `action.close`, `action.reject`, etc..., so that they can be passed around easier, like:

``` javascript
import promiseAjax from 'somewhere'

function searchAction (query) {
  return function (action) {
    action.open()
    promiseAjax.get('/search', query).then(action.close, action.reject)
  }
}
```

nhunzaker:
Going to close this. I can't figure out a way to do it without hindering perf.

nhunzaker:
Opening this up again. Let's analyze the perf and figure out how to make it work. This would be a big ergonomics win.

cc @greypants

nhunzaker:
Landed in `microcosm@12.7.0-alpha.4`

PR #298, opened 2017-04-14
Action status methods are autobound

nhunzaker:
Action status methods, like `action.resolve()`, or `action.reject()` are now auto-bound. This means that you can pass them directly into a callback, and the scope of the method will remain as the action. 

This is particularly useful when working with AJAX libraries. For example, when working with `superagent`:

```javascript
import superagent from 'superagent'

function getPlanets () {
  return action => {
    let request = superagent.get('/planets')

    request.on('request', action.open)
    request.on('progress', action.update)

    request.then(action.resolve, action.reject)
  }
}
```

I struggled with adding this for a long time. Auto-binding generally increases memory usage and adds extra instantiation overhead. However, using a getters to lazy-generate the functions seems to do the trick. 

I also added a strict-mode warning for when you try to complete an action that has already finished:

screen shot 2017-04-14 at 7 56 16 am

Related issues:
https://github.com/vigetlabs/microcosm/issues/144

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,31 @@
   change event, even if state didn't change.
 - The first argument of `repo.push` is passed into the `open` state of
   actions that return promises.
+- Action status changing methods are auto-bound, and will warn when a
+  completed action attempts to move into a new state (strict mode only)
+
+### Autobound action status methods
+
+Action status methods, like `action.resolve()`, or `action.reject()`
+are now auto-bound. This means that you can pass them directly into a
+callback, and the scope of the method will remain as the action. This
+is particularly useful when working with AJAX libraries. For example,
+when working with `superagent`:
+
+```javascript
+import superagent from 'superagent'
+
+function getPlanets () {
+  return action => {
+    let request = superagent.get('/planets')
+
+    request.on('request', action.open)
+    request.on('progress', action.update)
+
+    request.then(action.resolve, action.reject)
+  }
+}
+```
 
 ## 12.6.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,34 @@
 - Action status changing methods are auto-bound, and will warn when a
   completed action attempts to move into a new state (strict mode only)
 
-### Autobound action status methods
+### Auto-bound action status methods
 
-Action status methods, like `action.resolve()`, or `action.reject()`
-are auto-bound. This means that you can pass them directly into a
-callback, and the scope of the method will remain as the action. This
-is particularly useful when working with AJAX libraries. For example,
-when working with `superagent`:
+Action status methods like `action.resolve()` and `action.reject()`
+are auto-bound. They can be passed directly into a callback without
+needing to wrap them in an anonymous function.
+
+This is particularly useful when working with AJAX libraries. For
+example, when working with `superagent`:
+
+Instead of:
+
+```javascript
+import superagent from 'superagent'
+
+function getPlanets () {
+  return action => {
+    let request = superagent.get('/planets')
+
+    request.on('request', (data) => action.open(data))
+    request.on('progress', (data) => action.update(data))
+    request.on('abort', (data) => action.cancel(data))
+
+    request.then((data) => action.resolve(data), (error) => action.reject(error))
+  }
+}
+```
+
+You can do:
 
 ```javascript
 import superagent from 'superagent'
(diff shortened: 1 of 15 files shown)


src/action.js
```diff
@@ -305,8 +223,114 @@ inherit(Action, Emitter, {
...
+   * Cancelled state. The action was halted, like aborting an HTTP
+   * request.
+   * @param {*} [nextPayload]
+   */
+  cancel: {
+    get () {
+      return warnOrUpdate(this, 'cancel', true)
+    }
+  }
+
```
nhunzaker: This is a little goofy, but it essentially defines a getter that returns either a warning or an auto-bound action update function for a particular status. I want to move to ES6 classes, where I could just define `get cancel ()` more easily.

src/action.js
```diff
@@ -305,8 +223,114 @@ inherit(Action, Emitter, {
...
+ * @param {string} method
+ * @return A function that will warn developers that they can not
+ * update this action
+ */
+function createCompleteWarning (action, method) {
+  return function () {
+    console.assert(false, `Action "${action.command.name || action.type}" is already in ` +
+                   `the ${action.status} state, it will not change. Calling ` +
+                   `${method} will not change it.`)
+  }
```
nhunzaker: I know this will catch bugs. I'm curious how annoying it'll be though. Let's find out :)

greypants: why `console.assert(false, "the string")` vs `console.warn("the string")`? wider browser support or something?

nhunzaker: We should warn. Good call.

nhunzaker: Done with 36d74a6

rollup.config.js
```diff
@@ -19,7 +20,17 @@ const config = {
...
+  config.plugins.push(
+    strip({
+      debugger: true,
+      functions: ['console.*'],
+      sourceMap: true
+    })
+  )
+}
+
+if (process.env.MINIFY) {
```
nhunzaker: Had to update the rollup config here. I don't want warnings to hit the main build. Strict mode only

CHANGELOG.md
```diff
@@ -9,6 +9,32 @@
   Microcosm classes.
 - Fixed a bug where the first action dispatched would always fire a
   change event, even if state didn't change.
+- Action status changing methods are auto-bound, and will warn when a
+  completed action attempts to move into a new state (strict mode only)
+
+### Autobound action status methods
+
+Action status methods, like `action.resolve()`, or `action.reject()`
+are now auto-bound. This means that you can pass them directly into a
```
efatsi: feels like we should avoid phrases like `are now auto-bound` in documentation, and just say `are auto-bound`

nhunzaker: Okay. Sure thing, 👍 . FWIW, I removed the "now" from the Action docs.

nhunzaker: Is auto-bound a too jargony too?

efatsi: i don't really get it but not sure how more experienced js-devs would feel

CHANGELOG.md
```diff
@@ -9,6 +9,32 @@
...
+callback, and the scope of the method will remain as the action. This
```
efatsi: this is the last confusing thing for me in these docs: 
```
and the scope of the method will remain as the action
```

greypants: Scope means what the value of `this` will be inside the function when called. When binding a function with `.bind`, the first argument is the scope:

`someFunction.bind(scope, param1, param2)`

CHANGELOG.md
greypants: naw, I think it's useful to say.

CHANGELOG.md
nhunzaker: I think rewriting this to `and the value of `this` will remain the action` would clear this up quite a bit.

efatsi: thanks for clearing things up for me. i think either option sounds good, i know that scope is an integral js concept. the only thing i'd suggest then is to kill the use of the word `remain`. For someone who doesn't know about how things used to work prior to this PR it only confiscates what you're trying to say.

CHANGELOG.md
```diff
@@ -11,6 +11,53 @@
...
+    let request = superagent.get('/planets')
+
+    request.on('request', action.open)
+    request.on('progress', action.update)
+    request.on('abort', action.cancel)
+
+    request.then(action.resolve, action.reject)
+  }
+}
+```
```
nhunzaker: @efatsi did some 💄. Is this better, or worse?

efatsi: nice, explains what those words mean quite well

Issue #296, opened 2017-04-12
Pass along data to action.open callback when returning a promise


greypants:
When actions return a promise, It would be convenient if callbacks registered with `action.open` received the original payload. The main use case being for performing optimistic updates before hearing back from the server.

```js
function updateFoo(data) {
  return ajax.patch('foo', data)
}

repo.push(updateFoo, { foo: 'bar'})

// ...elsewhere in a domain or effect:
register() {
  return {
    [updateFoo.open]: data => console.log(data) // undefined instead of { foo: 'bar'}
  }
}
```

Here's where the work needs to be done: https://github.com/vigetlabs/microcosm/blob/master/src/coroutine.js#L51-L60

nhunzaker:
This is in the current alpha, if you'd like to give it a spin:

`yarn add microcosm@12.7.0-alpha.4`

greypants:
It's working great! A+

PR #299, opened 2017-04-14
Pass first argument of action parameters as open state of promises

nhunzaker:
Name says it all

```javascript
function getPlanet (id) {
  return fetch(`/planets/${id}`).then(res => res.json())
}

let action = repo.push(getPlanet, 'mars')

action.onOpen(data => console.log(data))
// 'mars'

action.onDone(data => console.log(data))
// { id: 'mars', name: 'Mars' }
```

Related issues:
https://github.com/vigetlabs/microcosm/issues/296

### DIFF
--- a/src/coroutine.js
+++ b/src/coroutine.js
@@ -39,7 +39,7 @@ function processGenerator (action, body, repo) {
  * the body of their associated command.
  */
 export default function coroutine (action, command, params, repo) {
-  let body = command.apply(action, params)
+  let body = command.apply(null, params)
 
   /**
    * Provide support for Promises:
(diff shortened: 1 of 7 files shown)


src/coroutine.js
```diff
@@ -38,7 +38,9 @@ function processGenerator (action, body, repo) {
  * Coroutine is used by an action to determine how it should resolve
  * the body of their associated command.
  */
-export default function coroutine (action, body, repo) {
+export default function coroutine (action, command, params, repo) {
+  let body = command.apply(action, params)
```
efatsi: noticed this went from `command.apply(null, params)` to `command.apply(action, params)`, why is that?

nhunzaker: Very good catch. No reason. This would have been a bug.

nhunzaker: Fixed with 766312e

nhunzaker:
@greypants You can test this in `microcosm@12.7.0-alpha.2`. Could you give it a shot?

PR #300, opened 2017-04-18
Add `get` note for use without optional fallback.

cwmanning:
Clarify basic failure case with a small snippet and a test. @nhunzaker this might be overkill for something so simple, but it seemed like it might be useful enough to call out.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build tools
- [x] Documentation
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing
- [ ] Commits reference open issues


test/unit/utils/get.test.js
```diff
@@ -46,4 +46,10 @@ describe('Utils.get', function () {
     expect(fallback).toBe(true)
   })
 
+  it('returns undefined if the fallback is undefined', function () {
+    let fallback = get(null, null)
+
+    expect(fallback).toBe(undefined)
+  })
```
nhunzaker: What do you think about re-writing this as `fallback = get({}, 'key')`. Or do you think we should cover both cases?

cwmanning: sounds good to me, will update shortly.

nhunzaker:
Looks great! Thanks!

PR #301, opened 2017-04-19
Add prettier

leobauza:

- [ ] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactor
- [x] Build tools
- [x] Documentation
- [ ] Other, please describe:

- [ ] Yes
- [x] No

n/a

- [x] All tests are passing

There was no issue set for this.

**Additional Information**

I definitely don't know about this @nhunzaker... but:

- Added prettier details on CONTRIBUTING.md (TL;DR run `make pretty` or `yarn run pretty`)
- I think we need to come up with more rules we want to add to the `eslint --fix` part of this
- I am particularly bother by not being able to line up `import .... from`'s and variable assignments
- Trailing comas?
- It's pretty fun to write super messy code and have it magically fixed though so 👍 to that

🌮


nhunzaker:
Awesome job. This is super cool. I'm surprised by how few changes there actually are.

> Added prettier details on CONTRIBUTING.md (TL;DR run make pretty or yarn run pretty)

👍. Though I'd be okay with just `yarn run pretty`, even though it's not quite as fun. Could we add a CI check [here](https://github.com/vigetlabs/microcosm/blob/master/circle.yml#L18)? Maybe do the following:

```
- yarn lint # could run prettier --check and eslint
- yarn test:cov
```

> I think we need to come up with more rules we want to add to the eslint --fix part of this

Just curious -- what is eslint fixing?

> I am particularly bother by not being able to line up import .... from's and variable assignments

I used to be a fan of this, but I've watched every instance of it slowly devolve into something like:

```
import React                     from 'react'
import Presenter                 from 'microcosm/addons/presenter'
import HolyMolySpecificComponent from './holy-moly-specific-component'
```

Which I don't believe is more readable, and the diffs on Github are a distraction in code reviews. 

> Trailing comas?

I'm personally not a fan of this, but others like it. Stuff like this shouldn't block progress on auto-formatting, but if we were to cast a vote, I'd say no trailing commas. The project currently does not use trailing commas, what do you think about nixing it and circling back when we've talked about it as a group? Or... if you don't feel like it, or really want trailing commas, let's just move forward with it. 🤷‍♂️ 

> It's pretty fun to write super messy code and have it magically fixed though so 👍 to that

Totally. What a time to be alive!

leobauza:
Yeah I was kind of surprised. The other project I am working on changed dramatically. But it had `flow` so my function parameters were wild. 

> Just curious -- what is eslint fixing?

Any eslint rule that has the little wrench next to it [ie. curly](http://eslint.org/docs/rules/curly) will be fixed automatically when `eslint --fix` is run. So after prettier does it's thing if our eslint rules disagree with something prettier does and its "fixable" it will format it the way eslint has it.

> Tailing comas?

Cool...I guess the beauty of it is that it doesn't matter, we just change the rule and run prettier on everything and no more trailing comas...or...trailing comas...or back and forth forever.

> import...froms

Yes. Thank you, it ends up looking weird 90% of the time unless you set some nifty webpack aliases to make all your paths resolve better.

But...I don't like this:

```javascript
let foo = "foo"
let fooer = "fooer"
// is there a rule to format it:
let foo     = "foo"
let fooer = "fooer"
```

Though maybe the same `import...from` argument applies to this. I just know people will have more of a problem with this one.

> CI check 

Will add, prob reach out to you to make sure I get what you mean. I think I do.

examples/canvas/app/boot.js
```diff
@@ -21,11 +21,11 @@ repo.addDomain('pixels', {
...
     return update(state, [x, y], this.rotateHue)
   },
-  register() {
+  register () {
     return {
-      [advance]: this.advance
+      [advance]: this.advance,
     }
-  }
+  },
```
cwmanning: can trailing commas be configured? this stands out to me over most of the changes so far

nhunzaker:
> Though maybe the same import...from argument applies to this. I just know people will have more of a problem with this one.

To me, it's the same argument. I can use my alignment macro in Atom or Emacs to align things up quickly, but I don't like how it creates a situation where you're incentivized to name things for their word fit, rather than their clarity. It's a shame, because it _does_ look really nice, but it just causes paralysis when writing variables, and clutters up diffs in code review.

examples/chatbot/app/views/parts/message.js
```diff
@@ -2,25 +2,25 @@ import React from 'react'
 
 class Message extends React.PureComponent {
   static defaultProps = {
-    user    : '',
-    message : '',
-    pending : false,
-    time    : new Date()
+    user:    '',
+    message: '',
```
cwmanning: I feel like aligned objects holds up about as well as `import` blocks. If this is coming from ESLint, and we're voting, can I vote to forego object key alignment altogether?

```js
{
  user: '',
  message: '',
  pending: false,
  time: new Date()
}
```

nhunzaker: 👍

examples/canvas/app/boot.js
leobauza: Yeah, we can remove the coma dangle through `prettier` or through `eslint --fix`. I will make that change

examples/chatbot/app/views/parts/message.js
leobauza: 👍

greypants: I'm willing to give it up 👍

CONTRIBUTING.md
```diff
@@ -43,6 +43,40 @@ Documentation found on the [Microcosm site](http://code.viget.com/microcosm) is
...
+or
+
+```bash
+make pretty
+```
+
+These will first run `prettier` to format our code and then run `eslint --fix` to make additional style changes and fixes.
+
+### Things to note about the prettier/eslint combo
+
```
leobauza: @nhunzaker this is new. While changing the npm script for `lint` I ran into some problems which led me to change some things. I summarized in this bit of the docs.

leobauza:
Another thing I wanted to flag for discussion was [JSX bracket spacing](https://github.com/prettier/prettier/issues/1097):

- if we want linting to include a check that prettier has been run then we have to have it in the only format prettier allows ie `{likeThis}` which is uglier than `{ withRoomToBreathe }`
- if we *don't* care about checking that prettier has been run in our linting then we can use `eslint --fix` to fix that.

I can't find an easy way to run a script that will check the combination of both rules. Since prettier has less overrides we have to default to it's settings :man_shrugging:

CONTRIBUTING.md
```diff
@@ -43,6 +43,40 @@ Documentation found on the [Microcosm site](http://code.viget.com/microcosm) is
...
+// eslint --fix would do this
+'I am using single quotes because it\'s eslint --fix'
+```
+
+We are letting this happen for two reasons:
+
+1. Write whatever you want, prettier will fix it, that's the point
+2. Our `yarn run lint` task would error out if we did a `eslint --fix`. There is no way to turn change prettier's behaviour but we can turn off the offending eslint rule
+
```
leobauza: @nhunzaker this is new. While changing the npm script for `lint` I ran into some problems which led me to change some things. I summarized in this bit of the docs.

cwmanning:
@leobauza I'm in the camp of one of the comments on the JSX thread:

> {expression} in JSX is similar to (expresson) in JS. Along this line of argument, there shouldn't be spaces around expression surrounded by {}.

That being said, I wouldn't lose any sleep over either spacing preference. Probably a decision for @nhunzaker.

cwmanning:
@leobauza this looks great, btw! thanks for working on it.

nhunzaker:
@cwmanning I will lose no sleep either. I like no-spaces, but I don't really care if Prettier is doing a good job of keeping line lengths to a minimum.

tommymarshall:
This is looking super clean. Well done, @leobauza !

PR #303, opened 2017-04-19
Handle spaces in key paths, increase key path test coverage

nhunzaker:
Pretty cut and dry. I'm working on some stuff that uses key paths and I noticed that spaces aren't allowed. This PR allows for change subscriptions like:

```javascript
repo.on('lists, items', callback)
// or
repo.on('lists,items', callback)
```

Not a new feature, it's just natural to add a space after a comma, so let's make sure we cover it.

### DIFF
--- a/src/key-path.js
+++ b/src/key-path.js
@@ -25,7 +25,7 @@ export function castPath (value) {
     return []
   }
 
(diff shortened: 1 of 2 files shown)


efatsi:
👍

Issue #305, opened 2017-04-19
Move all constructors to classes


nhunzaker:
- [ ] Replace constructors with classes
- [ ] Remove inherit utility

PR #309, opened 2017-04-24
Convert Action into an ES6 class

nhunzaker:
This is a part of getting Microcosm ready to move over to Flow.

Related issues:
https://github.com/vigetlabs/microcosm/issues/305

### DIFF
--- a/src/action.js
+++ b/src/action.js
@@ -1,38 +1,83 @@
+/**
+ * @fileoverview Actions encapsulate the process of resolving an
+ * action creator. Create an action using `Microcosm::push`:
+ */
+
 import Emitter from './emitter'
 import tag from './tag'
-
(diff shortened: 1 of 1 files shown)


src/action.js
```diff
@@ -1,38 +1,83 @@
...
+    this.timestamp = Date.now()
+    this.children = []
+
+    if (status) {
+      console.assert(this[status], 'Unexpected task status ' + status)
+      this[status]()
+    }
+  }
+
+  get type() {
```
leobauza: Keep in mind this regarding getters and setters with flow: https://github.com/facebook/flow/issues/2826

Not saying we shouldn't have them, just that I know it's a thing.

nhunzaker: Good call. This should be safe. Sounds like we just need to set an option.

PR #310, opened 2017-04-25
More class transfer

nhunzaker:
Migrates a few more classes. `Emitter` gets inherited in a couple of places, this ended up being a little more than I thought I would need to chew off. Otherwise the Es6 class check fails the build with something like "Can not call class constructor as function"


efatsi:
Looks like a good amount of furniture rearranging, anything in particular you want eyes on?

nhunzaker:
@efatsi Hmm. That's a good question. This is definitely furniture moving. Prep work for adding Flow. Do you think it would be better to just push this straight to master?

efatsi:
i'd vote yay

nhunzaker:
👍

src/microcosm.js
```diff
@@ -8,82 +8,84 @@ import CompareTree from './compare-tree'
...
+ */
+const DEFAULTS = {
+  maxHistory: 0,
+  parent: null,
+  batch: false
+}
 
-  let options = merge(Microcosm.defaults, this.constructor.defaults, preOptions)
+/**
+ * @fileoverview The Microcosm class provides a centralized place to
```
leobauza: Is this prepping for JSDoc as well?

nhunzaker: Yep

nhunzaker: Where it makes sense, anyway. Trying not to go too hog wild.

PR #314, opened 2017-04-26
Add "What you get" section to the README

nhunzaker:
I want to hit harder on what you get with Microcosm as early as possible.

Am I missing anything? Is this clear? 

It would also be cool to convert these points into tutorials, or explain them later in the readme. 

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ optimistic updates with ease.
 - A central place to keep track of all application state
 - The ability to [split up state](./docs/api/microcosm.md#fork) in
   large applications while still sharing common data
-- [Painless optimistic updates](./docs/api/history.md)
+- [Painless optimistic updates](./docs/recipes/ajax.md)
 - [A way to schedule work](./docs/api/actions.md) that understands
   Promises out of the box and moves through consistent states
 - [Opinions about architecture](./docs/guides/architecture.md)
--- a/README.md
+++ b/README.md
@@ -17,8 +17,7 @@ optimistic updates with ease.
   large applications while still sharing common data
 - [Painless optimistic updates](./docs/api/history.md)
 - [A way to schedule work](./docs/api/actions.md) that understands
-  Promises out of the box and moves through consistent states no
-  matter what work is being done.
+  Promises out of the box and moves through consistent states
 - [Opinions about architecture](./docs/guides/architecture.md)
 - [Addons for decoupling business and presentation logic](./docs/api/presenter.md)
 - 5kb gzipped (~15kb minified)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,7 +5,22 @@
 [![npm](https://img.shields.io/npm/v/microcosm.svg?maxAge=2592000)](https://www.npmjs.com/package/microcosm)
 [![npm](https://img.shields.io/npm/dm/microcosm.svg?maxAge=2592000)](https://www.npmjs.com/package/microcosm)
 
-Microcosm is a state management tool for [React](https://github.com/facebook/react) (and similar libraries). Keep track of user actions, cancel requests, and perform optimistic updates with ease.
+Microcosm is a state management tool
+for [React](https://github.com/facebook/react) (and similar
+libraries). Keep track of user actions, cancel requests, and perform
+optimistic updates with ease.
```
greypants: why the hard line breaks here?

README.md
```diff
@@ -5,7 +5,22 @@
...
+
+- A central place to keep track of all application state
+- The ability to [split up state](./docs/api/microcosm.md#fork) in
+  large applications while still sharing common data
+- [Painless optimistic updates](./docs/recipes/ajax.md)
+- [A way to schedule work](./docs/api/actions.md) that understands
+  Promises out of the box and moves through consistent states
+- [Opinions about architecture](./docs/guides/architecture.md)
+- [Addons for decoupling business and presentation logic](./docs/api/presenter.md)
+- 5kb gzipped (~15kb minified)
```
greypants: When talking about Microcosm, I've been asked by a couple Redux users, "Are actions asynchronous?" (they're hoping the answer is yes). Wonder if it's worth mentioning that they can be out of the box (Redux requires additional middleware as I understand it).

greypants: That's basically of covered by "A way to schedule work" - but since I got the same question phrased from two different people ("Are actions asynchronous?"), thought it might be work using that language.

greypants: Though I still like "A way to schedule work" too. It's more descriptive and less jargony.

README.md
nhunzaker: Emacs thing. Haha. I could probably disable it.

README.md
mackermedia: @greypants & @nhunzaker : Yes. With Redux I've seen https://github.com/gaearon/redux-thunk used to make actions asynchronous. Even then, we struggled quite a bit working with Promises and Redux's `dispatch` to make it so we could simply "do something after an action completes" in noob speak. I think it's definitely worth highlighting this as a strong point for microcosm out of the box.

nhunzaker: 👍. Making a note to work on this when we get a chance on Friday morning.

nhunzaker:
Took a second pass. What do you think?

mackermedia:
:+1: I like!

PR #315, opened 2017-04-28
Add flow config, flow definitions for Emitter and Utils

nhunzaker:
Adds Flow. Starts preliminary flow types for Emitter and the utilities.


src/emitter.js
```diff
@@ -1,22 +1,23 @@
...
+ * @flow
  * @fileoverview Emitter is an abstract class used by a few other
  * classes to communicate via events
  */
 
 import { isFunction } from './utils'
 
+type EventName = string
+type EventPayload = Array<any>
+type EventCallback = (...payload: EventPayload) => {}
```
leobauza: Wondering if we shouldn't pull all these type definitions into a `flow` folder or a `flow-typed` folder. Like [this](https://flow.org/en/docs/libdefs/creation/) suggests (I think)  or the way [vue.js does it](https://github.com/vuejs/vue/tree/dev/flow).

src/emitter.js
```diff
@@ -126,7 +118,7 @@ class Emitter {
   /**
    * Remove all events for a given scope
    */
-  _removeScope(scope) {
+  _removeScope(scope: any) {
```
leobauza: Just pointing out `any` is unsafe. Is there a good reason for using it vs mixed (I think) https://flow.org/en/docs/types/any/

src/utils.js
```diff
@@ -1,9 +1,17 @@
-import { castPath } from './key-path'
+/**
+ * @flow
+ */
```
leobauza: Wondering if you know anything about flow working together with JSDocs? Also, can we just use `// @flow` for brevity and to not confuse the JSDocs writing? Maybe that's a silly thing, but thoughts?

src/emitter.js
nhunzaker: I think think all of these just need to be mixed

src/utils.js
nhunzaker: I know DocumentationJS could generate JSDoc meta data from Flow types. My initial thoughts were to do something like:

```
/**
 * @fileoverview ....
 * @flow
 */
```

src/emitter.js
nhunzaker: Sure. Let's give it a shot.

src/utils.js
leobauza: ah, gotcha. Yeah that'll work better.

nhunzaker:
I'm going to close this out for now. Let's get jsdoc in place, then let's take a pass with flow.

PR #322, opened 2017-05-12
Add jsdoc

leobauza:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build tools
- [x] Documentation
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing

**Additional Information**

I wanted to put this out there to do a gut check that this is the direction we want to go with docs. This is the start of moving towards JSDocs.

What I have so far:

- A custom JSDoc template
- Ability to create versioned docs based on `package.json`
- A homepage matching the existing homepage
- Some sample tutorials to show how that might work
- Some of the documenation for the `microcosm` class showing how that might work
- A script to run the docs locally `jsdoc/scripts/serve` a script to create the docs `yarn docs`

I've talked to @nhunzaker and I think the best way to do this is to have this running parallel to the current docs until we are ready to fully move to JSDoc. For now that means:

- There is a `new_site` directory where the JSDocs are compiled
- There is a `jsdoc` directory that contains all the configs, templates, tutorials, and scripts for JSDoc stuff.
- There is a new task that can be run with `yarn docs` that will generate the docs at the top level of `new_site` and a nested folder for the versioned docs.
- There is a script in `jsdoc/scripts` that will let you run the JSDocs locally on `localhost:4000` (undocumented for now)

If this is the direction we'd like to go then I can start working on moving all the docs into JSDoc comments on `src` and organizing the tutorials (guides and stuff).

The idea would be to merge this so that new stuff being written doesn't get too out of sync with the effort to move the documentation into source files. Then eventually kill the current `site` and `docs` folder and make `new_site` the new `site`. 

@efatsi we can talk about whether you'd want what you are working on to be the front page of the new docs. It shouldn't be a problem if we want it to live with everything else here.


leobauza:
Forgot to mention that Ideally I would eventually write the `Microcosm` JSDoc template as a node_module so that we could just install it that way instead of it living on the repo. But I am currently working on it so it's in the repo right now.

jsdoc/scripts/serve
```diff
@@ -0,0 +1,36 @@
...
+//       console.error(stdout)
+//     }
+//
+//     bs.reload()
+//   })
+// }
+
+// bs.watch('../**/*.md').on('change', make)
+// bs.watch('../templates/**/*').on('change', make)
+// bs.watch('../tutorials/**/*').on('change', make)
```
nhunzaker: Should these be removed?

nhunzaker:
Awesome work!

PR #324, opened 2017-05-17
Adds repo.all for parallel actions. Sugar in yield action format.

nhunzaker:
This commit adds the ability to batch together a group of actions under a parent "group" action. If all of the actions complete or cancel, the group resolves. Otherwise the action is categorically rejected.

This surfaces itself through a new Microcosm method: `repo.all([...actions])`: 

```javascript
let group = repo.all([
  repo.push(actionOne),
  repo.push(actionTwo)
])

group.onDone(function () {
  console.log('hurrah!')
})
```

Additionally, generator actions can yield an array:

```javascript
function fetchUsers(usersList) {
  return function*(repo) {
    let requests = usersList.map(repo.prepare(fetchUser))

    yield requests
  }
}
```

- [x] Feature

- [x] No

**Related issues:**

- https://github.com/vigetlabs/microcosm/issues/229

### DIFF
--- a/src/action.js
+++ b/src/action.js
@@ -163,7 +163,7 @@ class Action extends Emitter {
   link(actions) {
     let outstanding = actions.length
 
-    const onDone = () => {
+    const onResolve = () => {
       outstanding -= 1
 
       if (outstanding <= 0) {
@@ -172,8 +172,8 @@ class Action extends Emitter {
     }
 
     actions.forEach(action => {
-      action.onDone(onDone)
-      action.onCancel(onDone)
+      action.onDone(onResolve)
+      action.onCancel(onResolve)
       action.onError(this.reject)
     })
 
--- a/src/action.js
+++ b/src/action.js
@@ -155,6 +155,32 @@ class Action extends Emitter {
   }
 
   /**
+   * Set up an action such that it depends on the result of another
+   * series of actions.
+   * @param {Array.<Action>} actions
+   * @return {this}
+   */
+  link(actions) {
+    let outstanding = actions.length
+
+    const onDone = () => {
+      outstanding -= 1
+
+      if (outstanding <= 0) {
+        this.resolve()
+      }
+    }
+
+    actions.forEach(action => {
+      action.onDone(onDone)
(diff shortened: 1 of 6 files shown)


src/action.js
```diff
@@ -155,6 +155,32 @@ class Action extends Emitter {
...
   /**
+   * Set up an action such that it depends on the result of another
+   * series of actions.
+   * @param {Array.<Action>} actions
+   * @return {this}
+   */
+  link(actions) {
+    let outstanding = actions.length
+
+    const onDone = () => {
```
efatsi: #super-nitpick would `onResolve` be a better name here?

nhunzaker: Yes.

efatsi:
`repo.all`, would `repo.sequence` be more clear? also opens up paradigm for `repo.parallel`?

efatsi:
i only have renaming suggestions, code looks really good, quite impressive how small the footprint of this PR is

nhunzaker:
@efatsi `repo.all` is inspired by `Promise.all`, which sort of does the same thing only for Promises. This behavior is essentially the same as a `repo.parallel`. The actions fire off at the same time and `repo.all` waits for them all to complete.

I was thinking about adding `repo.sequence`, but we have the `yield` syntax for generators.

efatsi:
@nhunzaker ah, thanks for the clarification. as an outsider i would prefer to see terminology like `parallel` and `sequence` (highlighted by the fact that i misunderstood `all`'s functionality out of the gate), as they more accurately describe their functionality, but if this is a strong convention then I won't stand in it's way.

And yes while we have the `yield` generator syntax for sequential actions, I think it would be nice if there was a higher level syntax sugar for running sequential actions, and then you could dip down into the `yield` syntax if you needed more complex functionality (referencing previous actions in later ones).

thoughts?

nhunzaker:
Totally. I think that's a great idea.

nhunzaker:
@efatsi Thanks for the update! I really like `repo.parallel` a lot. Any outstanding concerns with merging?

efatsi:
shipit!

PR #325, opened 2017-05-17
Sequential actions

efatsi:
Similar to @nhunzaker's recent work on `repo.all` (or `repo.parallel`), this adds `repo.sequence` as a way to queue up a group of actions that are resolved/rejected as one unit. Since we already have support for returning a generator in an action, this is really just syntax sugar in front of that work:

```js
let group = repo.sequence([
  repo.prepare(actionOne),
  repo.prepare(actionTwo)
])

group.onDone(function () {
  console.log('hurrah!')
})
```

In that example, the return value of `actionOne` will be passed in as an available argument to `actionTwo`. Another option for kicking this off is which bypasses that argument passing:

```js
let group = repo.sequence([
  () => repo.push(actionOne),
  () => repo.push(actionTwo)
])
```

- [x] Feature

- [x] No

- [x] All tests are passing

**Related Issue**

- https://github.com/vigetlabs/microcosm/issues/229


nhunzaker:
Wooah. Very cool. I can't look at it in depth right at the moment, but this is super awesome.

docs/api/actions.md
```diff
@@ -159,6 +159,26 @@ is rejected or cancelled with the same payload.
...
 action will be the resolved payload of the final action.
 
+#### Yielding Actions in Parallel
+
+By yielding an array of actions, you can wait for multiple actions to
+complete before continuing:
+
+```javascript
+function getUser (id) {
+  return fetch(`/users/${id}`)
```
nhunzaker: I just pushed a thing to master. `fetch` has a couple of problems, like it doesn't reject the promise it returns if it receives a 400. I've updated my examples to read:
```
import axios from 'axios'

function getUser (id) {
  // axios returns a promise
  return axios(`/users/${id}`)
}
```

docs/api/actions.md
```diff
@@ -159,6 +159,26 @@ is rejected or cancelled with the same payload.
...
+
+function getUsers (ids) {
+  return function * (repo) {
+    yield ids.map(id => repo.push(getUser, id))
+  }
+}
+```
+
+If all actions resolve or cancel, the generator sequence
+continues.
```
nhunzaker: 👍 👍 👍

docs/api/microcosm.md
```diff
@@ -329,6 +329,57 @@ fork.addDomain('page', PaginatedPeople)
...
+`actionOne` passed in as the first argument. If this is
+undesirable, you can bypass this functionality like so:
+
+```javascript
+let group = repo.sequence([
+  () => repo.push(actionOne),
+  () => repo.push(actionTwo)
+])
+```
+
```
nhunzaker: This is fantastic, :+1:.

docs/api/microcosm.md
```diff
@@ -329,6 +329,57 @@ fork.addDomain('page', PaginatedPeople)
...
+The returned value of any given action will be passed in as an
+additional parameter to the following action. So in our above
+example, `actionTwo` would be called with the return value of
+`actionOne` passed in as the first argument. If this is
+undesirable, you can bypass this functionality like so:
+
+```javascript
+let group = repo.sequence([
+  () => repo.push(actionOne),
+  () => repo.push(actionTwo)
```
nhunzaker: Passing the payload of the last action to the next is easy to miss. Is there a quick way for us to show that?

package.json
```diff
@@ -9,7 +9,7 @@
     "build": "make all",
     "lint": "prettier --single-quote --semi false -l \"{src,test,examples}/**/*.js\" && eslint \"{src,test,examples}/**/*.js\"",
     "pretty": "prettier --single-quote --semi false --write \"{src,test,examples}/**/*.js\" && eslint --fix \"{src,test,examples}/**/*.js\"",
-    "test": "npm run lint && NODE_ENV=test jest --maxWorkers=2",
+    "test": "NODE_ENV=test jest --maxWorkers=2",
```
nhunzaker: Agreed. Let's make sure this still runs on CI though.

src/action.js
```diff
@@ -155,6 +155,32 @@ class Action extends Emitter {
...
+    actions.forEach(action => {
+      action.onDone(onResolve)
+      action.onCancel(onResolve)
+      action.onError(this.reject)
+    })
+
+    return this
+  }
+
+  /**
```
nhunzaker: I've seen this before... Does this branch just need to be updated?

src/microcosm.js
```diff
@@ -260,6 +260,22 @@ class Microcosm extends Emitter {
...
+    let payload
+
+    let sequence = () => function * () {
+      for (var i = 0; i < prepared.length; i++) {
+        payload = yield prepared[i](payload)
+      }
+    }
+
+    return this.push(sequence)
+  }
```
nhunzaker: I really like how clean this is, but if we use generators inside of Microcosm, we'll need to pull in the [regenerator runtime](https://github.com/facebook/regenerator) that makes them possible in older browsers.

It's like 600 bytes, but I'd rather let people patch their own run time in. How could we write this so that it does not depend on generators? Could we call `reduce` on prepared?

test/unit/microcosm/prepare.test.js
```diff
@@ -9,4 +9,13 @@ describe('Microcosm::prepare', function() {
...
+
+  it('does not push the action by default', function() {
+    const repo = new Microcosm()
+    const action = jest.fn()
+
+    repo.prepare(action, 1, 2)
+
+    expect(action).not.toBeCalled()
+  })
 })
```
nhunzaker: 👍

test/unit/middleware/generator-middleware.test.js
```diff
@@ -186,4 +186,53 @@ describe('Generator Middleware', function() {
...
+          yield [repo.push(add, 1), repo.push(add, 1)]
+        }
+      }
+
+      repo.push(testReturnValue).onDone(() => {
+        expect(repo).toHaveState('count', 2)
+      })
+    })
+  })
 })
```
nhunzaker: :+1:

docs/api/actions.md
efatsi: huh, not actually sure where this update came from. perhaps something you had in the `repo-all` branch that is no longer there. either way, will update to axios and keep.

docs/api/microcosm.md
efatsi: thoughts on updating and adding a bit of explanation?

```javascript
let group = repo.sequence([
  ()        => repo.push(actionOne),
  (payload) => repo.push(handlesActionOneResponse),
  ()        => repo.push(isolatedAction),
])
```

So here, the response of `actionOne` would be passed in to the second action, but the response of that would be ignored when the third `isolatedAction` is called.

package.json
efatsi: dernit, didn't mean to commit this. the linter was blocking me from running the tests while i was figuring out how to get a generator to `yield` from within a `map` (you can't`) so i pulled it out. I'll throw it back in as I don't have enough context to suggest an alternative.

src/microcosm.js
efatsi: ahh, that i'm not sure about. the reason this is so clean is because the work for walking through actions sequentially is already done in `coroutine::processGenerator`. something like

nhunzaker: Yeah, we'd need to make a callback that processed the outstanding work. Maybe like `.parallel`, only daisy chain

efatsi: Tempted to put this PR on the back burner for now. The ability to run with sequential actions is available for those who desire.

PR #326, opened 2017-05-18
Presenter model

nhunzaker:
This PR breaks out the Presenter view modeling logic into a separate `Model` class. Additionally, it adds support for [Observables](https://github.com/tc39/proposal-observable).

[Observables](https://github.com/tc39/proposal-observable) are a solution for tracking changes over time that have become popular in a lot of JavaScript communities. 

I want to be able to set up a Presenter model binding that can describe a value over time. This could be a query string, a binding to repo state, or progress updates. Observables solve that particular problem:

```javascript
class Count extends Presenter {
  getModel () {
    return {
      count: Observable.of(1, 2, 3)
    }
  }
  render () {
    return <p>{this.model.count}</p>
  }
}

DOM.render(<Count />, document.body) // Displays 1...2...3, but probably just 3
```

I am also curious about how we could interop with actions, assuming they implemented the correct interface:

```javascript
class Count extends Presenter {
  getModel () {
    return {
      count: this.repo.push(getCount)
    }
  }
  render () {
    const { count } = this.model

    if (count.status === 'open') {
      return <p>Loading</p>
    }

    return <p>count.payload</p>
  }
}

DOM.render(<Count />, document.body) // Displays "Loading".... "[number]"
```

But this is highly experimental.

I think Observables they're worth paying attention to. I would love to lean on them more inside of Microcosm. Reading more about them, I believe they describe exactly what I want for model bindings, and perfectly describe the `Action` lifecycle (not status + payload, but their lifecycle within History). They would clean up internals quite a bit, and I'm starting to see how they could make interoperability with other libraries easier.

---

- [x] Feature

- [x] No

- [x] All tests are passing

**Additional Information**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   represents a group of actions processing in parallel.
 - The action generator form may now `yeild` an array. This produces
   the same behavior as `repo.parallel`
+- `Presenter::getModel` assignments
+  accept [Observables](https://github.com/tc39/proposal-observable).
 
 ## 12.8.0
 


src/addons/model.js
```diff
@@ -0,0 +1,123 @@
...
+
+      next = set(next, key, value)
+    }
+
+    if (last !== next) {
+      this.value = next
+      this._emit('change', this.value)
+    }
+
+    return next
```
efatsi: why does this return anything?

src/addons/model.js
```diff
@@ -0,0 +1,123 @@
...
+
+    this.subscriptions[key] = next
+
+    if (last) {
+      last.unsubscribe()
+    }
+  }
+
+  /**
+   * @param {Object} bindings An set of key/value pairs for building a model
```
h0tl33t: `A set` instead of `An set`

src/addons/model.js
```diff
@@ -0,0 +1,123 @@
...
+   * Update a specific model key. Emits a change event
+   * @param {string} key
+   * @param {*} value
+   */
+  set(key, value) {
+    let next = set(this.value, key, value)
+
+    if (this.value !== next) {
+      this.value = next
+      this._emit('change', this.value)
```
efatsi: you have `this._emit('change', this.value)` here as well as in `compute`, if i'm reading this correctly, will that result in n+1 separate change emissions if there are n changes to `this.value`? And, will the last emission (in `compute`) be redundant since the the previous change event would have emitted the most up-to-date `this.value`?

those questions aside, my understanding of why you've done this is so that observables have a way to emit change events when something occurs there. am i on base there?

src/addons/model.js
```diff
@@ -0,0 +1,123 @@
...
+
+      if (isObservable(binding)) {
+        this.track(key, binding)
+      } else {
+        this.bindings[key] = binding
+      }
+    }
+
+    this.compute()
+  }
```
efatsi: a couple questions here:

1. why does this need to be called every time `updateModel` happens? my understanding is that `bindings` don't change from one update to the next, am i wrong there?
2. if the answer to 1 is compelling, then I think `bind` isn't the best terminology here, `bind`, to me at least, implies a one time setup deal, where as something like `evaluate` might be more applicable

src/addons/model.js
```diff
@@ -0,0 +1,123 @@
...
+  constructor(repo, scope) {
+    super()
+
+    this.repo = repo
+    this.scope = scope
+    this.bindings = {}
+    this.subscriptions = {}
+    this.value = {}
+
+    this.repo.on('change', this.compute, this)
```
efatsi: another thing i'm curious about, it looks like `compute` could be triggered from a couple different angles. you're triggering a call here, but also in the flow of `Presenter::componentWillUpdate -> _prepareModel -> updateModel -> model.bind -> this.compute`. Would this result in duplicitous `compute` calls every time the repo state changed?

nhunzaker: I'll check.

src/addons/model.js
nhunzaker: Thank you!

src/addons/model.js
nhunzaker: Originally for convenience. It's not necessary. I'll nix this.

src/addons/model.js
nhunzaker: No. Compute is not called duplicitously. At least for the test cases that I have. I made a few more, but then realized that the same use cases are covered in the `Efficiency` test suite section of `presenter.test.js`.

src/addons/model.js
nhunzaker: 1. When a presenter gets new props or state, we invoke `getModel`. This is so that you can create bindings specific to a particular prop. This happens quite a bit for URL parameters:

```javascript
getModel (props) {
  const { params } = props

  return {
    user: state => state.users.find(user => user.id === params.id
  }
}
```

getModel is not recalculated if repo state changes. Only presenter props/state.

2. I don't like the name `bind` either. This is essentially updating the model with a fresh set of data subscriptions. Any suggestions?

---

Beyond that, this method of building a view model comes from Redux's `Connect/Provide` pattern. For a while, I've felt like it's not very efficient. You execute the model callbacks every time, relying on immutable data to make it fast. This isn't good enough, a lot of the times we run `array.filter` or something, which always returns a new array, causing an unnecessary render. You can avoid this for a primary data calculation, but you almost always get into trouble if you have some sort of secondary calculation. 

So something to chew on. This is part of the reason I want to move towards observables.

src/addons/model.js
nhunzaker: That's correct. Observables operate on a single key, so we need to update that key as they change, and bubble that change up to the Presenter. 

So yes. This ends up with n+1 in the following case:

```
getModel () {
  return {
    a: state => state.prop,
    b: new Observable(observer => {
      this.repo.on('change', state => observer.next(state.otherProp)
    })
  }
}
```

In this case, we'd get two updates. One option would be to batch, sort of like what we do for Microcosm change events. The other would be to call `compute` when an observable changes, executing the function callbacks if state has changed.

Tough problem :).

src/addons/model.js
efatsi: Thanks a lot for clearing this up for me. Given the context I think `bind` is suitably accurate.

src/addons/model.js
nhunzaker: Sorry. I should also clarify. This will send out two updates, but the Presenter assigns state when the model changes. So, although there are two change events, ideally the Presenter only fires one change because the setState callbacks are batched.

efatsi: Aha. The batching of setState callbacks is all I need to hear. 👍

efatsi:
Looks like all my flag raising have been false alarms, so with those fully squelched, 👍

nhunzaker:
@efatsi still really good concerns. Thanks for the awesome review!

PR #327, opened 2017-05-18
Add documentation to presenter methods

nhunzaker:
Adds examples for all presenter methods

---

- [x] Documentation

- [x] No

### DIFF
--- a/docs/api/presenter.md
+++ b/docs/api/presenter.md
@@ -1,31 +1,25 @@
 # Presenter
 
 1. [Overview](#overview)
-2. [Computed Properties](#computed-properties)
+2. [Track changes and compute values](#track-changes-and-compute-values)
 3. [Receiving Actions](#receiving-actions)
 4. [API](#api)
 
 ## Overview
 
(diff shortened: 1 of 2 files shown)


docs/api/presenter.md
```diff
@@ -156,23 +156,89 @@ function StepperForm ({ count }) {
...
+  getModel (props) {
+    const { planetId } = props
+
+    return {
+      planet: state => state.planets.find(planet => planet.id === planetId)
+    }
+  }
+  // These are the next props and state
+  update (repo, props, state) {
+    if (props.planetId !== this.props.planetId)
```
efatsi: so `props` is the next set of props, and `this.props` here is the last set, which will soon be set to equal `props`?

efatsi:
nice 👍

docs/api/presenter.md
nhunzaker: Yep. Hmmm. You know what. This should be `nextProps`, `nextState`

PR #329, opened 2017-05-23
Microcosms can be register to actions

nhunzaker:
Microcosms can now declare a custom register and getInitialState method. In this way, they behave exactly like domains.

This does not allow Microcosms to be added as domains to other Microcosms. That will require some work. This is intended to make it easier to introduce Microcosm, and for concise examples.

```javascript
const ADD = 'ADD'

class Repo extends Microcosm {
  getInitialState() {
    return { count: 0 }
  }

  add(state, n) {
    return update(state, 'count', value => value + n)
  }

  register() {
    return {
      [ADD]: this.add
    }
  }
}

let repo = new Repo()

repo.push(ADD, 2)
repo.push(ADD, 2)

expect(repo).toHaveState('count', 4)
```

---

- [x] Feature

- [x] No

- [x] All tests are passing

### DIFF
--- a/src/domain-engine.js
+++ b/src/domain-engine.js
@@ -1,16 +1,16 @@
 import MetaDomain from './meta-domain'
 import getRegistration from './get-registration'
 import { get, set, createOrClone } from './utils'
-import { castPath } from './key-path'
+import { castPath, getKeyString } from './key-path'
 
 class DomainEngine {
   /**
    * @param {Microcosm} repo
    */
   constructor(repo) {
-    this.repo = repo
-    this.domains = []
     this.registry = {}
+    this.repo = repo
+    this.domains = [[[], this.repo]]
 
     // All realms contain a meta domain for basic Microcosm operations
     this.add([], MetaDomain)
@@ -76,14 +76,26 @@ class DomainEngine {
     return next
   }
 
+  supportsKey(key) {
+    if (key in this.repo.state) {
+      return true
+    }
+
+    return this.domains.some(entry => getKeyString(entry[0]) === key)
+  }
+
   sanitize(data) {
+    let repo = this.repo
+    let parent = repo.parent
     let next = {}
 
-    for (var i = 0, len = this.domains.length; i < len; i++) {
-      let [key] = this.domains[i]
+    for (var key in data) {
+      if (parent && parent.domains.supportsKey(key)) {
+        continue
+      }
 
-      if (key.length) {
-        next = set(next, key, get(data, key))
+      if (this.supportsKey(key)) {
+        next[key] = data[key]
       }
     }
 
--- a/src/domain-engine.js
+++ b/src/domain-engine.js
@@ -10,10 +10,19 @@ class DomainEngine {
   constructor(repo) {
     this.registry = {}
     this.repo = repo
-    this.domains = [[[], this.repo]]
+    this.domains = []
 
     // All realms contain a meta domain for basic Microcosm operations
     this.add([], MetaDomain)
+
+    // Microcosms can register actions
+    this.addRepo(repo)
+  }
+
+  addRepo(repo) {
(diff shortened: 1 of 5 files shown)


src/domain-engine.js
```diff
@@ -1,16 +1,16 @@
...
 class DomainEngine {
   /**
    * @param {Microcosm} repo
    */
   constructor(repo) {
-    this.repo = repo
-    this.domains = []
     this.registry = {}
+    this.repo = repo
+    this.domains = [[[], this.repo]]
```
efatsi: just starting to review this PR so i don't know what this is doing or why it's changed but it looks pretty gross

src/microcosm.js
```diff
@@ -173,7 +173,7 @@ class Microcosm extends Emitter {
    * Generates the starting state for a Microcosm instance. This is the result of dispatching `getInitialState` to all domains. It is pure; calling this function will not update state.
    */
   getInitialState() {
-    return this.initial
+    return this.initial == null ? {} : this.initial
```
efatsi: is this equivalent to `this.initial || {}`?

src/domain-engine.js
efatsi: ah, so `domains` is an array where every item is an array with the first element being the keypath for the domain, and the second being the domain (or Microcosm now) that registers to actions

src/microcosm.js
nhunzaker: No. If initial were 0 or false, it would incorrectly return an object.

src/domain-engine.js
nhunzaker: Yeah. This is gross a bit. It would be great to just use a `Map` here, even if we wrote a light version for the sake of this project.

efatsi:
cool stuff, 2 things I'm curious about:
- would you mind explaining the motivation behind the change to `sanitize()`
  - also, digging into when `sanitize` is invoked, is it possible that it's invoked multiple times? looking at the following execution path:
    - [Microcosm::patch](https://github.com/vigetlabs/microcosm/blob/3de6df6c668054bffd555aa4d8be531ba35758ec/src/microcosm.js#L315-L317)
    - [Lifecycle::PATCH -> sandbox](https://github.com/vigetlabs/microcosm/blob/3de6df6c668054bffd555aa4d8be531ba35758ec/src/lifecycle.js#L18) calls `repo.domains.sanitize`
    - [MetaDomain::patch](https://github.com/vigetlabs/microcosm/blob/3de6df6c668054bffd555aa4d8be531ba35758ec/src/meta-domain.js#L29-L33) is registered to PATCH action and also calls `repo.domains.sanitize`
- What are your thoughts on simply merging `repo.register` within the body of `MetaDomain::register`? So instead of pushing `[[], this.repo]` onto the list of Domains, the action registration functionality is just handled by the MetaDomain. The fact that a Microcosm instance would live side by side with all of it's Domains seems a little funky to me conceptually, and more like a convenient hack to broaden the reach of the DomainEngine's functionality. Also, pushing it on the stack even when it has no `register` functionality seems wasteful.

nhunzaker:
@efatsi I like the idea of merging register into the meta domain. The problem is that scope then becomes the meta domain, instead of the Microcosm, so if you call another method, like `this.something`, the method is accessed on MetaDomain, not the microcosm.

As for sanitize. I'll dig in here. Basically, we want to prevent forks from patching new state into their parents. Like if you do:

```
let repo = new Microcosm()
let fork = repo.fork()

fork.patch({ foo: 'bar' })
```

`repo.state` should not include `foo`.

efatsi:
@nhunzaker gotchya. well then my only suggestion would be to wrap the `[[], this.repo]` push in it's own method, and possibly only invoke it if you can determine that the repo has implemented it's own register.

otherwise, 👍

nhunzaker:
@efatsi Done. :)

PR #330, opened 2017-05-23
Do not warn in strict mode when changing a completed action

nhunzaker:
This allows for use cases like, "Cancel this action, but only if it hasn't finished yet.":

```javascript
class PlanetsList extends Presenter {
  getModel () {
    return {
      planets: state => state.planets
    }
  }
  setup (repo) {
    this.action = repo.push(getPlanets)
  }
  teardown () {
    // We no longer care about this
    this.action.cancel()
  }
  // ...
}
```

---

- [x] Feature

- [x] No

- [x] All tests are passing


efatsi:
👍

PR #333, opened 2017-05-24
Implement toJSON for Action and History

nhunzaker:
This allows serialization of history for developer tools. This lets you do:

```javascript
JSON.stringify(repo.history) 
// { size: 3, root: { id: 0, status: 'resolved', payload: true, children: [...] }
```

This is for:

https://github.com/vigetlabs/microcosm-devtools

### DIFF
--- a/src/action.js
+++ b/src/action.js
@@ -290,6 +290,16 @@ class Action extends Emitter {
       }
     }
   }
+
+  toJSON() {
+    return {
+      id: this.id,
+      status: this.status,
+      type: this.type,
+      payload: this.payload,
+      children: this.children
+    }
+  }
 }
 
 /**
--- a/src/history.js
+++ b/src/history.js
@@ -345,6 +345,18 @@ class History extends Emitter {
       cursor = cursor.parent
     }
   }
+
+  /**
+   * Serialize history into JSON data
+   */
+  toJSON() {
+    return {
+      head: this.head.id,
+      root: this.root.id,
+      size: this.size,
+      tree: this.root
+    }
+  }
 }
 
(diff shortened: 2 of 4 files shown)


src/action.js
```diff
@@ -290,6 +290,16 @@ class Action extends Emitter {
...
     }
   }
+
+  toJSON() {
+    return {
+      id: this.id,
+      status: this.status,
+      type: this.type,
+      payload: this.payload,
+      children: this.children
```
h0tl33t: Are `payload` and `children` serializable?

src/history.js
```diff
@@ -297,6 +297,13 @@ class History extends Emitter {
 
     this.size = size
   }
+
+  toJSON() {
+    return {
+      size: this.size,
+      root: this.root
```
efatsi: if the action tree is multi-branched, should there be something here that denotes the head?

src/action.js
nhunzaker: Children is. JSON.stringify recursively calls `toJSON` on properties (and so on). Payload actually is an interesting one. It might not be serializable.

src/history.js
nhunzaker: Ah good call. We probably want an identifier for the head.

nhunzaker:
@efatsi Made a few updates and added some test coverage. I feel like this is probably going to need to change to fit the needs of the dev tools, but it at least illustrates the technique.

src/action.js
nhunzaker: But it would be serializable if the payload implemented toJSON

PR #335, opened 2017-05-24
Fix bugs in history.remove

efatsi:

- [x] Bugfix

- [x] No

- [x] All tests are passing

**Additional Information**

This PR fixes two bugs:
- If you remove an action which has no `next` reference, History assumed you were removing the head action and automatically set the head to the action's parent. This ensures that the head is reset only if the action being removed was the existing head.
- If you remove the head action, set the head to the removed action's parent

### DIFF
--- a/src/history.js
+++ b/src/history.js
@@ -179,22 +179,29 @@ class History extends Emitter {
       return
     }
 
-    let next = action.next
+    // cache linking references and activeness
     let parent = action.parent
+    let next = action.next
+    let wasActive = this.isActive(action)
 
     this.clean(action)
 
+    // if there are no more actions left, we're done
     if (this.size <= 0) {
       this.begin()
       return
-    } else if (action === this.head) {
+    }
+
+    // reset head/root references if necessary
+    if (action === this.head) {
       next = this.head = parent
     } else if (action === this.root) {
       this.root = next
     }
 
-    if (!action.disabled) {
-      this.reconcile(this.head)
+    // reconcile history if action was in active history branch
+    if (wasActive && !action.disabled) {
+      this.reconcile(next)
     }
   }
 
@@ -296,6 +303,14 @@ class History extends Emitter {
     }
 
     this.size = size
+  }
+
+  /**
+   * Determine if provided action is within active history branch
+   * @param {Action} action
+   */
+  isActive(action) {
+    return this.toArray().indexOf(action) >= 0
   }
 }
 
--- a/src/history.js
+++ b/src/history.js
@@ -193,8 +193,8 @@ class History extends Emitter {
       this.root = next
     }
 
-    if (!action.disabled && next) {
-      this.reconcile(next)
+    if (!action.disabled) {
+      this.reconcile(this.head)
     }
   }
 
--- a/src/history.js
+++ b/src/history.js
@@ -187,7 +187,7 @@ class History extends Emitter {
     if (this.size <= 0) {
       this.begin()
       return
-    } else if (action === this.head && !next) {
+    } else if (action === this.head) {
       next = this.head = parent
     } else if (action === this.root) {
       this.root = next
--- a/src/history.js
+++ b/src/history.js
@@ -187,13 +187,13 @@ class History extends Emitter {
     if (this.size <= 0) {
       this.begin()
       return
-    } else if (!next) {
+    } else if (action === this.head && !next) {
       next = this.head = parent
     } else if (action === this.root) {
       this.root = next
     }
 
-    if (!action.disabled) {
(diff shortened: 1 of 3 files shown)


src/history.js
```diff
@@ -187,13 +187,13 @@ class History extends Emitter {
...
+    } else if (action === this.head) {
       next = this.head = parent
     } else if (action === this.root) {
       this.root = next
     }
 
-    if (!action.disabled) {
+    if (!action.disabled && next) {
       this.reconcile(next)
     }
```
efatsi: i feel like this probably needs a reconsideration as well, my guess is that we should only call `reconcile` with `this.head` if the head has changed, but wanted to run a gut check on that intuition before pursuing.

test/unit/history/remove.test.js
```diff
@@ -129,8 +123,6 @@ describe('History::remove', function() {
       let two = history.append(function two() {}, 'resolve')
       history.append(function three() {}, 'resolve')
 
-      history.archive()
-
```
efatsi: not sure what these were doing, specs pass without them so hopefully this is okay...

nhunzaker: Forcing a reconciliation.

nhunzaker:
Dang. Excellent work, 👍

src/history.js
```diff
@@ -187,14 +187,14 @@ class History extends Emitter {
...
-    } else if (!next) {
+    } else if (action === this.head) {
       next = this.head = parent
     } else if (action === this.root) {
       this.root = next
     }
 
     if (!action.disabled) {
-      this.reconcile(next)
+      this.reconcile(this.head)
```
nhunzaker: We need to reconcile history at the point of change. What happens in the following scenario?

```
[a] -> [b] -> [c]
```

1. Remove a

With this change, history would reconcile at _`c`_, when it needs to be _`_b`. Do I have that right?

efatsi: Not sure if I'm thinking about this too much, but is there a lot of logic that needs to go into this line right here? I''m imagining this like I picture git branches and interactive rebases that remove commits. So for a focused history branch I see removing an action causing a reconcile from that action's parent, and for an unfocused branch i guess i picture no reconciliation, but one would need to occur when you switched from branch to branch starting at the last shared action. Is this can of worms something we need to worry about?

nhunzaker: Ahhhh. I see now. Okay. It sounds like we need to know if the action is within the current branch. We could do that two ways:

```
// get the active branch
let isActive = this.toArray().indexOf(next) >= 0

// Traverse to the head
let isActive = false
let cursor = next
while (cursor) {
  if (cursor === head) {
    isActive = true
    break;
  }
  cursor = cursor.next
}
```

The later is probably more performant, but I'm not sure how much it really matters.

efatsi: okay, so thinking I'll update this code here for removing active/non-active actions, and then also update the `checkout` logic to reconcile from the most recent shared action - that sound right?

here, in the case of an active action, would reconciliation need to start at the parent of the removed action?

efatsi: ignore last question there, understanding how a microcosm reacts to the `update` emission from history
```js
let next = this.recall(action.parent, this.initial)
```

efatsi:
@nhunzaker pushed up https://github.com/vigetlabs/microcosm/pull/335/commits/ea698ceac2929c7cbb8b7fc787264944b262a3b2 to reconcile on the right action node, in the process of also refactoring `checkout` to refactor on the shared root (and a couple of other things), will spin up a new PR for that one

src/history.js
```diff
@@ -297,6 +304,14 @@ class History extends Emitter {
 
     this.size = size
   }
+
+  /**
+   * Determine if provided action is within active history branch
+   * @param {Action} action
+   */
+  isActive(action) {
+    return this.toArray().indexOf(action) >= 0
```
nhunzaker: Just a stray, mostly inactionable thought, but I wonder if we could design action ids to be able to tell if one action proceeds another in the same branch of history. 

Stray thought. This is great, and very easy to understand.

test/unit/history/isActive.test.js
```diff
@@ -0,0 +1,36 @@
...
+    expect(history.isActive(two)).toEqual(true)
+    expect(history.isActive(five)).toEqual(true)
+    expect(history.isActive(six)).toEqual(true)
+  })
+
+  it('returns false for actions not in the active branch', function() {
+    expect(history.isActive(three)).toEqual(false)
+    expect(history.isActive(four)).toEqual(false)
+  })
+})
```
nhunzaker: 👍

PR #336, opened 2017-05-24
Ef/checkout refactor

efatsi:

- [x] Bugfix
- [x] Refactor

- [x] No

- [x] All tests are passing

**Additional Information**

In preparation for a refined debugger that allows you to hop around the history tree, this PR addresses a couple existing issues:
1. Calling `checkout` only reset the `next` ref of the direct parent of the action you were checking out. If you checked out a branch that was more than one action away from the most recent shared action, that would leave the `next` chain pointing down the wrong branch
2. If you remove an action from within an inactive branch, we won't reconcile on that action (per PR #335). This means that checking out an action further in that branch would have an invalid state. This PR addresses that by determining the most recent action shared between the current head and the newly desired head, and reconciling on that (after properly hooking up the `next` chain)

TODO:
Realizing now that `toggle` needs a similar treatment that `remove` just underwent, or we can pull a cop out and disable toggling of inactive actions.

### DIFF
--- a/src/history.js
+++ b/src/history.js
@@ -179,22 +179,29 @@ class History extends Emitter {
       return
     }
 
-    let next = action.next
+    // cache linking references and activeness
     let parent = action.parent
+    let next = action.next
+    let wasActive = this.isActive(action)
 
     this.clean(action)
 
+    // if there are no more actions left, we're done
     if (this.size <= 0) {
       this.begin()
       return
-    } else if (action === this.head) {
+    }
+
+    // reset head/root references if necessary
+    if (action === this.head) {
       next = this.head = parent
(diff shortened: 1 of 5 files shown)


src/history.js
```diff
@@ -297,6 +312,39 @@ class History extends Emitter {
...
+  sharedRoot(action) {
+    let cursor = action
+
+    while (cursor) {
+      if (this.isActive(cursor)) {
+        return cursor
+      }
+      cursor = cursor.parent
+    }
+  }
```
efatsi: not sure if we have a pattern in place for internal or private methods? `isActive` and `sharedRoot` just exist to help abstract functionality out to help keep the public methods cleaner

Issue #334, opened 2017-05-24
Document history events


nhunzaker:
History emits events. This is how a Microcosm knows to update its snapshots of state:

https://github.com/vigetlabs/microcosm/blob/master/src/microcosm.js#L110-L123

We should document this. Maybe even have a graphic here:

http://code.viget.com/microcosm/api/history.html

PR #337, opened 2017-05-24
Add documentation for History reconciliation and event emission

efatsi:

- Documentation

Re: https://github.com/vigetlabs/microcosm/issues/334

### DIFF
--- a/docs/api/history.md
+++ b/docs/api/history.md
@@ -2,6 +2,7 @@
 
 1. [Overview](#overview)
 2. [API](#api)
+2. [Events](#events)
 
 **Note: This is a work in progress document and feature. History has
 been a private API for a very long time. We're still working through
@@ -67,6 +68,27 @@ In the example above, the following tree would be produced:
 branch is used to determine state. Here, a Microcosm's state will be
 the result of dispatching `root`, `one`, and `three` to Domains.
 
+### Reconciling
+
+What makes History such a powerful feature of Microcosm is it's
+ability to walk through recorded actions whenever there is a change,
+similar to a rebase in git.
+
+Let's say we have four actions pushed to our History, and that three
+of the four are completed `(C)`, but the second action we pushed is
+taking a long time and is still open `(O)`.
+```
+[root] - [one](C) - [two](O) - [three](C) - [four](C)
+```
+
+When our `two` action here resolves or rejects, History will initiate
+a reconciling starting at `two`, and walking forward to `three` and
+then `four`. At each step, your Domains will recompute changes to the
+state based on the actions.
+
+This ensures that your application state is always accurate based on
+the order in which actions were triggered.
+
 ## API
 
 ### checkout(action)
@@ -93,7 +115,7 @@ calculate state by reconciling `root`, `one,` and `three`.
 
 Disable a group of actions:
 
-```
+```javascript
 let repo = new Microcosm({ maxHistory: Infinity })
 
 let one = repo.push(addUser)
@@ -109,7 +131,7 @@ existed.
 This flips the `disabled` state of each action provided. By executing
 toggle a second time, these actions will be re-enabled:
 
-```
+```javascript
 // Actions disabled in the prior example
 repo.history.toggle([ one, two ])
 // `one` and `two` have been re-enabled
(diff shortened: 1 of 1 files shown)


docs/api/history.md
```diff
@@ -190,3 +212,134 @@ repo.history.remove(two)
...
+
+repo.push(newAction)
+repo.push(newAction)
+repo.push(newAction)
+
+// Action: 1
+// Action: 2
+// Action: 3
+// Released!
+```
```
nhunzaker: Damn. This is awesome.

docs/api/history.md
```diff
@@ -67,6 +68,27 @@ In the example above, the following tree would be produced:
 branch is used to determine state. Here, a Microcosm's state will be
 the result of dispatching `root`, `one`, and `three` to Domains.
 
+### Reconciling
+
+What makes History such a powerful feature of Microcosm is it's
```
nhunzaker: its*

docs/api/history.md
```diff
@@ -67,6 +68,27 @@ In the example above, the following tree would be produced:
...
+
+Let's say we have four actions pushed to our History, and that three
+of the four are completed `(C)`, but the second action we pushed is
+taking a long time and is still open `(O)`.
+```
+[root] - [one](C) - [two](O) - [three](C) - [four](C)
+```
+
+When our `two` action here resolves or rejects, History will initiate
+a reconciling starting at `two`, and walking forward to `three` and
```
h0tl33t: a "reconciliation" starting at

docs/api/history.md
```diff
@@ -67,6 +68,27 @@ In the example above, the following tree would be produced:
...
+then `four`. At each step, your Domains will recompute changes to the
```
h0tl33t: Should this read "to their state" vs. "to the state"?

docs/api/history.md
```diff
@@ -190,3 +212,133 @@ repo.history.remove(two)
 
 repo.history.toArray() // [root, one, three]
 ```
+
+## Events
+
+History emits events any time something of interest happens. This is
+how Microcosm knows to update the state to accurately reflect what's
+going on given a sequence of actions and their varying statuses.
```
h0tl33t: Can drop "varying" IMO.

docs/api/history.md
```diff
@@ -190,3 +212,133 @@ repo.history.remove(two)
...
+```javascript
+const repo = new Microcosm()
+const history = repo.history
+
+history.on('append', callback)
+```
+
+### `off(event, callback)`
+
+Remove an event listener.
```
h0tl33t: To stay consistent with description of `on`, maybe change this to "Removes an event listener".

docs/api/history.md
```diff
@@ -2,6 +2,7 @@
 
 1. [Overview](#overview)
 2. [API](#api)
+2. [Events](#events)
```
mackermedia: 3. (instead of 2.)

docs/api/history.md
mackermedia: `its ability` instead of `it's ability`

docs/api/history.md
mackermedia: `walk forward` vs `walking forward`?

docs/api/history.md
```diff
@@ -190,3 +212,133 @@ repo.history.remove(two)
...
+### `off(event, callback)`
+
+Remove an event listener.
+
+```javascript
+// Remove a callback
+history.off('append', callback)
+```
+
+### Event Types
```
mackermedia: the previous `on` & `off` are at heading level `###`, then there's `### Event Types` at level three, and the following `### append` (+remove, update) are also all at level three. Should Event Types be at a different heading level? Maybe not?

docs/api/history.md
```diff
@@ -190,3 +212,133 @@ repo.history.remove(two)
...
+
+repo.push(newAction)
+
+// Action status: inactive
+// Action status: open
+// Action status: update
+// Action status: [resolve, reject, cancel]
+```
+
+Whenever an action status updates that precedes other actions in
```
mackermedia: this reads slightly off... maybe `When an action status updates and that action precedes other already updated/resolved actions in History,` ?

mackermedia:
👍

docs/api/history.md
efatsi: technically they're just things that mutate the repo's state, think i'll leave it at "the state"

docs/api/history.md
efatsi: Yeah I need like a 2.5 header since this is within `## Events` but above `### append`. Initially I had just a sentence here introing the types but felt I needed something more punchy. Open to other ideas but this is the best I can think of for now

docs/api/history.md
efatsi: thoughts on `Whenever an action that precedes other actions has a status update, ...`?

don't care about the status of the other actions

efatsi:
@nhunzaker @mackermedia @h0tl33t thanks for the feedback, ran with all the suggestions unless i specified otherwise

PR #340, opened 2017-06-02
Prevent case where Microcosm register method caused infinite loop

nhunzaker:
If a microcosm was added to a domain engine, the deserialize method was getting picked up. This commit changes the way microcosm registration happens to avoid an infinite loop.


mackermedia:
👍

PR #342, opened 2017-06-05
Move more query/node concerns to their associated modules

nhunzaker:
Microcosm has an internal tree for comparing data. This is so that you can do things like `repo.on('change:planets', (planets) => {})`. This commit reduces some responsbilities inside of the comparison tree itself so that individual nodes have greater encapsulation.

I've also added an efficiency for calculating keypaths that avoids re-parsing array-form keypaths. Paths that are already in the correct form are not additionally parsed. 

For example:

```javascript
let path = [['foo', 'bar']]
let parsed = castPath(path)

assert.equal(path, parsed)
```

### DIFF
--- a/src/compare-tree/index.js
+++ b/src/compare-tree/index.js
@@ -1,6 +1,6 @@
 import Node from './node'
 import Query from './query'
-import { getKeyPaths, getKeyString } from '../key-path'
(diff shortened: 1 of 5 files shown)


efatsi:
👍

PR #347, opened 2017-06-15
Add Flow

nhunzaker:
This commit adds FlowType to Microcosm. Additionally, it implements
types for the internal archive package.

Away we go.


src/archive.js
```diff
@@ -1,9 +1,16 @@
 /**
  * @fileoverview Keeps track of prior action states according to an
  * action's id
+ * @flow
  */
 
+import type Action from './action'
+
+type Payload = { [string]: * }
```
leobauza: Can you explain this bit to me? So you are defining a `Payload` type and it is an object that has...and then I don't know what the rest means.

src/archive.js
```diff
@@ -16,36 +23,45 @@ class Archive {
...
+    return typeof this.pool[action.id] !== 'undefined'
   }
 
   /**
    * Assign a new snapshot for an action
    * @param {Action} action Action for requested snapshot
    * @param {Object} snapshot
    */
-  set(action, snapshot) {
+  set(action: Action, snapshot: Payload) {
```
leobauza: You have `remove()` returning `void` but this unspecified. Does flow automatically assume `void` as a return value or should we add that here or take it off of `remove()`

nhunzaker: Ah, good catch. I believe it does. I think we should just omit `:void`.

src/archive.js
nhunzaker: Yeah. It is an object where the keys are strings and the values could be anything.

nhunzaker: Actually... That's pretty useless. I'll just reduce this down to Object.

Issue #346, opened 2017-06-14
Cleaning up the generator action interface


nhunzaker:
Generator actions are quickly becoming my favorite form of writing actions:

```javascript
function createUserFlow (params) {
  return function * (repo) {
    yield repo.push(validate, params)

    let user = yield repo.push(createUser, '/users')

    yield repo.push(navigate, '/users/' + user.id)
  }
}
```

This feels heavy. I wonder if we could clean this up by adding a layer of abstraction:

```javascript
// 1. A general "send" helper, where you don't need to know about the repo at all
import { send } from 'microcosm'

// 2. Actions themselves can be generators, not just returning a function
function * createUserFlow (params) {
  yield send(validate, params)

  let user = yield send(createUser, '/users')

  yield send(navigate, '/users/' + user.id)
}
```

Under the hood, `send` might do something like:

```
function send (action, ...params) {
  return repo => repo.push(action, ...params)
}
```

We'd pass `repo` to the action inside of Microcosm.

efatsi:
seems like a good idea to me

PR #348, opened 2017-06-16
RFC: Revise React Router example with some future plans

nhunzaker:
React Router v4 has a more component based API. It also makes it easier to work directly with history. This PR is an exploration of what future Microcosm apps might look like. 

### Flow

I've added FlowTypes. I want first class support for them. I still have to figure some stuff out for the `register()` method, which is quite dynamic, but I'm happy so far.

### Domains are responsible for CRUD

In the future, I would like domains to own management of their data concerns. They have their own actions. What we call actions now are more like controllers:

```javascript
/**
 * @flow
 */
import uid from 'uid'
import Domain from './domain'

export type List = {
  id?: string,
  name: string
}

class Lists extends Domain {
  static create(params): List {
    return { id: uid(), ...params }
  }

  static destroy(id: string) {
    return id
  }

  register() {
    return {
      [Lists.create]: this.add,
      [Lists.destroy]: this.remove
    }
  }
}

export default Lists
```

### Action based navigation

Effects can accept options. In this PR, I make a effect for routing. This lets us do navigation within actions:

```javascript
export function addList(params) {
  return function*(repo) {
    let list = yield repo.push(Lists.create, params)

    yield repo.push(visit, `/lists/${list.id}`)
  }
}
```

### Models

I want to reduce the logic inside of Presenters. Increasing the responsibility of actions is part of that, but I also am interested in how Presenters can compose model bindings together, pulling out a lot of that logic into discrete, easy to test, components:

```javascript
export class ListWithCounts implements Model {
  call(_presenter, state) {
    const { lists, items } = state

    return lists.map(function(list) {
      let count = items.filter(i => i.list === list.id).length

      return set(list, 'count', count)
    })
  }
}

class ListIndex extends Presenter {
  getModel() {
    return {
      lists: new ListsWithCounts()
    }
  }

  render() {
    const { lists } = this.model
    // ...
  }
}
```

## What's next

This is a work in progress. I don't expect this to merge. I would like to make a few improvements:

1. Add more sugar for sequential actions (#346)
2. Add more sugar for models. Could they use Observables? Could we use the Formula concept we've internally applied to a recent project?


examples/react-router/app/boot.js
```diff
@@ -1,19 +1,18 @@
...
 import { AppContainer } from 'react-hot-loader'
 import Repo from './repo'
 import Layout from './views/layout'
 
 const el = document.getElementById('app')
-const repo = new Repo({ maxHistory: Infinity })
-
-Debugger(repo)
+const browserHistory = createBrowserHistory()
+const repo = new Repo({ maxHistory: Infinity, browserHistory })
```
nhunzaker: By sending in the same history to our Repo and React Router, the Routing effect can manage navigation. This is also very easy to test:

```javascript
import { createMemoryHistory } from 'history'

let repo = new Repo({ browserHistory: createMemoryHistory() })

// .. do other things, and we don't need a real browser. We can
// also assert on history
```

examples/react-router/app/domains/domain.js
```diff
@@ -0,0 +1,28 @@
...
+  remove(items: Collection, unwanted: string) {
+    return items.filter(i => i.id !== unwanted)
+  }
+
+  removeBy(key: string) {
+    return (items: Collection, value: *) => {
+      return items.filter(item => item[key] !== value)
+    }
+  }
+}
```
nhunzaker: This is the beginnings of a basic domain for CRUD over a collection. I imagine this will evolve significantly.

PR #351, opened 2017-06-20
Add flow types to utilities. Optimize merge.

nhunzaker:
This commit adds flow types to the utilities module. It also uncovers an optimization where we can avoid returning a new object when the right-most value of a `merge` call is a superset. For example:

```javascript
let left = { green: true }
let right = { green: true, blue: true }
let answer = merge(left, right)

expect(answer).toBe(right)
```

We can just return the right-most value because it contains no differences with the left.

### DIFF
--- a/src/domain-engine.js
+++ b/src/domain-engine.js
@@ -129,7 +129,7 @@ class DomainEngine {
     return next
   }
 
-  dispatch(state: Object, action: Action) {
+  dispatch(state: Object, action: Action): Object {
     let handlers = this.register(action)
 
     for (var i = 0, len = handlers.length; i < len; i++) {
--- a/src/utils.js
+++ b/src/utils.js
@@ -39,33 +39,27 @@ export function clone(target: ?Object): Object {
  * @private
  */
 export function merge(): Object {
-  var copy = {}
-  var first = copy
-  var dirty = false
+  let copy = null
+  let subject = null
 
-  for (var i = arguments.length - 1; i >= 0; i--) {
-    var subject = arguments[i]
+  for (var i = 0, len = arguments.length; i < len; i++) {
+    copy = copy || arguments[i]
+    subject = subject || copy
 
-    if (isObject(subject) === false) {
-      continue
-    }
+    var next = arguments[i]
 
-    if (first === copy) {
-      first = subject
-    }
+    for (var key in next) {
+      if (copy[key] !== next[key]) {
+        if (copy === subject) {
+          copy = clone(subject)
+        }
 
-    for (var key in subject) {
-      if (!dirty) {
-        dirty = key in first === false
-      }
-
-      if (key in copy === false) {
-        copy[key] = subject[key]
+        copy[key] = next[key]
       }
     }
   }
 
-  return dirty ? copy : first
+  return copy || {}
 }
 
 /**
--- a/src/utils.js
+++ b/src/utils.js
@@ -161,7 +161,7 @@ export function isObject(target: *): boolean {
  * @return {boolean}
  * @private
  */
-export function isFunction(target: *): boolean {
+export function isFunction(target: any): boolean {
   return !!target && typeof target === 'function'
 }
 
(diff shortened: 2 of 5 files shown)


src/utils.js
```diff
@@ -33,36 +38,42 @@ export function clone(target) {
...
-        copy[key] = next[key]
+      if (key in copy === false) {
+        copy[key] = subject[key]
       }
     }
   }
 
-  return copy
+  return dirty ? copy : first
 }
```
efatsi: this is super cool, my only comment (as per usual), is that it's a bit tough to reason over.
- I think `first` is a misleading name since it's actually set to be the _last_ argument being passed in (i think). Hard to think of a better name though...
- `if (first === copy) { first = subject }` is the check that `first === copy` here the same as checking that `first` hasn't been set to something yet? If so, I think if it would be more clear if you initialized `var first = null` and checked it against null. Unless I'm missing something about the copy equality.
- The following example for the setting of `dirty` is much easier to reason about in my mind, no hard feelings but if you have similar thoughts, here's what my pass would be:
```js
if (!dirty && !(key in first)) {
  dirty = true
}
```

efatsi:
Couple comments, but 👍

src/utils.js
nhunzaker: I'm with you. Good thoughts. I'll take a second pass.

src/domain-engine.js
```diff
@@ -129,7 +129,7 @@ class DomainEngine {
     return next
   }
 
-  dispatch(state: Object, action: Action) {
+  dispatch(state: Object, action: Action): Object {
```
leobauza: I am wondering whether it would be a useful check to do something like https://flow.org/en/docs/types/generics/#toc-function-types-with-generics for this where `dispatch` takes a `state` which is of a generic type and returns the same generic type. Though I could be misunderstanding how that works...orrrrr that may not be useful

src/utils.js
```diff
@@ -166,7 +177,7 @@ export function isFunction(target) {
  * @return {boolean}
  * @private
  */
-export function isString(target) {
+export function isString(target: *): boolean {
```
leobauza: does the `*` mean `any`? or `mixed`? or just letting it infer? if the last one then is it needed?

src/utils.js
```diff
@@ -33,36 +38,42 @@ export function clone(target) {
...
-  let subject = null
+export function merge(): Object {
+  var copy = {}
+  var first = copy
+  var dirty = false
 
-  for (var i = 0, len = arguments.length; i < len; i++) {
-    copy = copy || arguments[i]
-    subject = subject || copy
+  for (var i = arguments.length - 1; i >= 0; i--) {
```
leobauza: why do you use `arguments` vs `...args` or something (I might be getting that wrong). Just curious, don't think you need a change or anything.

nhunzaker: I don't want to allocate an array to convert `arguments` into an `args` array.

nhunzaker:
@leobauza @efatsi I need to revise the merge stuff in here. I've been thinking about it more, and I think it's a bad idea to be hyper efficient about this stuff. For example: using `merge` to shallow copy an array

```javascript
let answer = merge({}, { foo: "bar" })
```

In this case, idiomatically I would expect to get a new object, and I might even take advantage of that to mutate the result. However with my new approach, we would just modify the second argument in place. We don't want that.

src/domain-engine.js
nhunzaker: Absolutely. This is a great idea.

nhunzaker: And really hard to figure out within the context of `get` and `set`. I'm making a note to circle back on this.

src/utils.js
nhunzaker: The existential type! It says "let flow infer it"

https://flow.org/en/docs/types/utilities/#toc-the-existential-type

But basically, any value is permissible here.

nhunzaker: Maybe this should just be `any`

src/domain-engine.js
leobauza: Is this why flow discourages `get` and `set`? or am I thinking of something else? Wondering if that's related.

nhunzaker: Ah, I meant the `get()` and `set()` data helpers.

PR #352, opened 2017-06-23
Add types for compare tree

nhunzaker:
This commit adds flow type coverage for CompareTree, Node, and Query

### DIFF
--- a/src/compare-tree/index.js
+++ b/src/compare-tree/index.js
@@ -1,28 +1,31 @@
+/**
+ * @flow
+ * @private
+ */
+
 import Node from './node'
 import Query from './query'
-import { getKeyString } from '../key-path'
+import { getKeyString, type KeyPath } from '../key-path'
 
 // The root key is an empty string. This can be a little
 // counter-intuitive, so we keep track of them as a named constant.
 const ROOT_PATH = ''
 
 class CompareTree {
-  /**
-   * @param {*} initial Starting state
-   */
-  constructor(initial) {
+  snapshot: Object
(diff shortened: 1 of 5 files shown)


src/compare-tree/index.js
```diff
@@ -1,28 +1,31 @@
...
-   */
-  constructor(initial) {
+  snapshot: Object
+  nodes: { [key: string]: Node }
+  queries: { [key: string]: Query }
+
+  constructor(initial: Object) {
     this.snapshot = initial
     this.nodes = {}
+    this.queries = {}
```
nhunzaker: I split the `Query` type into a separate map. This got me really deep flow coverage for a few cases. Still, I think it would be nice to eventually have `Node` and `Query` work from a common interface, if possible.

src/compare-tree/index.js
```diff
@@ -1,28 +1,31 @@
+/**
+ * @flow
+ * @private
+ */
+
 import Node from './node'
 import Query from './query'
-import { getKeyString } from '../key-path'
+import { getKeyString, type KeyPath } from '../key-path'
```
leobauza: is this importing a class and then turning it into a flow type from `key-path`...if so...TIL. Is this different than just importing the class (without saying `type`) and just using that as a type?

nhunzaker: KeyPath is a type alias for `string | Array<string>`. 

I believe the difference is that `type <module>` does not create a hard dependency, it only imports the type information. So if `getKeyString` weren't also imported here, the compare tree would not directly depend on the `key-path` module.

PR #354, opened 2017-06-28
Full flow coverage

nhunzaker:
Adds remaining flow coverage.

This is a pretty big change. Thank goodness for train rides to and from HQ 😅. Happy to break this down into smaller pieces. I'll make a few notes along the way too.


examples/react-router/app/domains/lists.js
```diff
@@ -1,22 +1,28 @@
...
     return lists.filter(i => i.id !== unwanted)
-  },
+  }
 
   register() {
     return {
-      [addList]: Lists.add,
-      [removeList]: Lists.remove
+      [addList.toString()]: this.add,
+      [removeList.toString()]: this.remove
```
nhunzaker: Flow does not like coercing these register methods to strings. So I've added toString here. *This is not awesome*. We might want to explore an alternative registration method that is easier to add type checking for.

This sort of thing will not be merged :).

flow/effect.js
```diff
@@ -0,0 +1,29 @@
...
+  teardown(repo: Microcosm): void,
+
+  /**
+   * Returns an object mapping actions to methods on the effect. This is the
+   * communication point between a effect and the rest of the system.
+   */
+  register(): {
+    [any]: (repo: Microcosm, action: Action) => void
+  }
+}
```
nhunzaker: @leobauza Curious if we could generate docs out of this.

src/action.js
```diff
@@ -27,122 +46,67 @@ class Action extends Emitter {
...
   }
 
-  /**
-   * Open state. The action has started, but has received no response.
-   * @param {*} [nextPayload]
-   */
-  get open() {
-    return createActionUpdater(this, 'open', false)
+  get open(): ActionUpdater {
+    return this.setState.bind(this, 'open')
```
nhunzaker: It turns out, `.bind` is faster than returning a new function every time. Also it made things cleaner.

src/addons/model.js
```diff
@@ -1,19 +1,20 @@
...
-import { set, Emitter } from '../microcosm'
+import { type Microcosm, Emitter, set } from '../microcosm'
 
-function isObservable(binding) {
+function isObservable(binding: *): boolean {
   return binding && typeof binding.subscribe === 'function'
 }
 
-function isCallable(binding) {
+function isCallable(binding: *): boolean {
```
nhunzaker: Using * means "let Flow infer it" This is super neat because you get really fantastic output from Flow while still being a little permissive about "This could literally accept anything"

src/coroutine.js
```diff
@@ -21,19 +25,21 @@ function processGenerator(action, body, repo) {
...
+  function progress(subAction: Action | Action[]): Action {
     if (Array.isArray(subAction)) {
-      subAction = repo.parallel(subAction)
+      return progress(repo.parallel(subAction))
     }
 
     console.assert(
       subAction instanceof Action,
-      `Iteration of generator expected an Action. Instead got ${subAction}`
+      `Iteration of generator expected an Action. Instead got ${typeof subAction}`
```
nhunzaker: Flow does not like implicit coercion, so I added a `typeof` here.

flow/effect.js
leobauza: yeah I am wondering that too (and thing's like `status` that I commented above

src/action.js
```diff
@@ -155,10 +119,8 @@ class Action extends Emitter {
   /**
    * Set up an action such that it depends on the result of another
    * series of actions.
-   * @param {Array.<Action>} actions
-   * @return {this}
    */
-  link(actions) {
+  link(actions: Action[]): this {
```
leobauza: is this the same as `Array<Actions>`...it is...cool TIL. This is shorter I should change my stuff to be like this. But `Array<Actions>` reads better to me. When we talk about how we will eventually do flow we should point this out and chose a standard.

src/addons/model.js
leobauza: but doesn't leaving it out let flow infer it also? I don't know what the difference is here?

nhunzaker: Huh. It does, but I can only do it for functions. If I attempt to do it for a class method, it yells at me. 

My preference is to be explicit. "I meant for this, I didn't forget". What do you think?

nhunzaker:
Neat. I'll add `eslint-plugin-flowtype-errors` to the project. I wonder how that interacts with standard flow plugins for editors (like nuclide for atom and flycheck for emacs).

src/addons/model.js
leobauza: makes sense 👍

PR #356, opened 2017-07-05
Simplify history.wait()

nhunzaker:
Now that we can group actions, [history.wait()](http://code.viget.com/microcosm/api/history.html#wait) can be simplified.

### DIFF
--- a/src/history.js
+++ b/src/history.js
@@ -132,30 +132,13 @@ class History extends Emitter {
    * Return a promise that represents the resolution of all actions in
    * the current branch.
    */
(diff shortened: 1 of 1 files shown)


efatsi:
a little dense, but clean 👍

nhunzaker:
@efatsi I'm always torn on that sort of stuff. Preference should go to those who didn't write the code. Would you like to see this broken out like:

```
let action = new Action('group')

action.link(this.toArray())

return action.then(...)
```

nhunzaker:
Decompressed!

PR #357, opened 2017-07-07
More efficient comparison

nhunzaker:
I noticed this working through some performance stuff on our Formula abstraction for a client project. We can reduce the amount of garbage created by checking for changes quite a bit. Additionally, we had no formal benchmark for measuring changes to the comparison tree.

This reduces memory consumption by:

- Avoiding extra parsing of change event handlers. `change:x,y` was getting parsed into `['x', 'y']` twice, instead of once
- Avoid an extra allocation building parameters for triggering events


bench/compare-tree-performance.js
```diff
@@ -0,0 +1,65 @@
...
+
+  repo.addDomain('pixels', {
+    getInitialState() {
+      return {}
+    },
+    rotateHue(n) {
+      return isNaN(n) ? 0 : n + 5
+    },
+    advance(state, { x, y }) {
+      return update(state, [x, y], this.rotateHue)
```
ipbrennan90: does giving update an array of keys update all keys to the passed value? I'm guessing this is the case but I didn't find anything in the docs.

nhunzaker: We should improve the docs on this. `get`, `set`, and `update` accept a key path. This could be:

1. A string: `a.b.c`
2. An array of string keys `['a', 'b', 'c']`

ipbrennan90:
👍

efatsi:
fwiw 👍

PR #358, opened 2017-07-07
added string keypath documetnation to immutability helpers that can u…

ipbrennan90:
- Documentation

- No

Just adding to the documentation for the immutability helpers to make it clear that get, set, and update can all take arrays of key strings or a string keypath.


docs/api/immutability-helpers.md
```diff
@@ -3,7 +3,7 @@
 1. [Overview](#overview)
 2. [API](#API)
 
-## Overview
+# Overview
```
nhunzaker: Could you keep this as an H2?

docs/api/immutability-helpers.md
```diff
@@ -58,10 +58,13 @@ let state = {
 let planets = get(state, 'planets')
 console.log(planets) // { earth, mars }
 
-// Or an array of keys to access a deeply nested value
+// an array of keys to access a deeply nested value
```
nhunzaker: Could you capitalize the a here?

nhunzaker: Hmmm. This is inconsistent. What if we just capitalized the first character of every comment in the file?

ipbrennan90: sounds good

PR #359, opened 2017-07-07
Actions that push from presenters should use same repo

nhunzaker:
This commit fixes an issue where actions, that float to the top of the presenter tree, are dispatched from the root repo, instead of the repo for the presenter that dispatched them. This causes an unexpected behavior where the root repo does not have state added by a domain within a fork.

Yes. This actually has happened on a client project:

```javascript
class Parent extends Presenter {
  setup(repo) {
    repo.addDomain('parent', {
      getInitialState() {
        return true
      }
    })
  }
}

class Child extends Presenter {
  setup(repo) {
    repo.addDomain('child', {
      getInitialState() {
        return true
      }
    })
  }
}

let action = function() {
  return function(action, repo) {
    expect(repo.state.child).toBe(true)
  }
}

let wrapper = mount(
  <Parent>
    <Child>
      <ActionButton action={action} />
    </Child>
  </Parent>
)

wrapper.find('ActionButton').simulate('click')
```

Each presenter fork adds a new domain. When you click the action button, it might be in a "sub-app". The expectation is that state from that "sub-app"'s fork should be present in the action.

### DIFF
--- a/src/addons/presenter.js
+++ b/src/addons/presenter.js
@@ -10,7 +10,7 @@
  */
 
 import React from 'react'
-import Microcosm, { get, merge, tag, getRegistration } from '../microcosm'
+import Microcosm, { merge, tag, getRegistration } from '../microcosm'
 import Model from './model'
 
 function passChildren() {
@@ -166,12 +166,14 @@ class Presenter extends React.PureComponent {
 class PresenterMediator extends React.PureComponent {
   static contextTypes = {
     repo: identity,
-    send: identity
+    send: identity,
+    parent: identity
   }
 
   static childContextTypes = {
     repo: identity,
-    send: identity
+    send: identity,
+    parent: identity
   }
 
   repo: Microcosm
@@ -197,7 +199,8 @@ class PresenterMediator extends React.PureComponent {
   getChildContext() {
     return {
       repo: this.repo,
-      send: this.send
+      send: this.send,
+      parent: this
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/addons/presenter.js
```diff
@@ -248,33 +251,44 @@ class PresenterMediator extends React.PureComponent {
...
   hasParent(): boolean {
-    // Do not allow transfer across repos. Check to for inheritence by comparing
-    // the common history object shared between repos
+    // No need to check if there isn't a parent context setup
+    if (!this.context.parent) {
+      return false
+    }
+
     return get(this.repo, 'history') === get(this.context, ['repo', 'history'])
   }
```
efatsi: can this function not just be something like:
```js
hasParent(): boolean {
  return !!this.context.parent
}
```

nhunzaker: There might be two separate histories, like if you mount one microcosm app into another (we hit this on iContact with the email editor and photo manager).

efatsi: so you're using history as a unique identifier of sorts? It's confusing to me to see history used in comparison here, doesn't seem like it has a logical reason to be in this area of code. But my understanding is that the only problem this is solving is the fact that an action should be caught by the closest repo available, and then passed up the parent chain if it doesn't catch it, perhaps i'm missing something about the problem.

src/addons/presenter.js
```diff
@@ -248,33 +251,44 @@ class PresenterMediator extends React.PureComponent {
     return this.model.value
   }
 
+  // Do not allow transfer across repos. Check to for inheritence by comparing
```
mackermedia: **to**

mackermedia:
👍

src/addons/presenter.js
nhunzaker: > my understanding is that the only problem this is solving is the fact that an action should be caught by the closest repo available

Precisely.

> so you're using history as a unique identifier of sorts? It's confusing to me to see history used in comparison here

Yes. We just need to prevent parent apps from stealing the actions of children. We may be able to tell by checking for the `repo` prop instead.

> doesn't seem like it has a logical reason to be in this area of code.

What do you recommend?

src/addons/presenter.js
nhunzaker: Thanks!

src/addons/presenter.js
nhunzaker: Yep. That's it:

8827ff7

efatsi:
looks great 👍

PR #360, opened 2017-07-10
Add revisions to actions

nhunzaker:
I'm working on a "revisions" feature to illustrate the timeline of an action. For that, I need to know prior revisions for an action:

screen shot 2017-07-10 at 8 09 12 am


test/unit/history/toJSON.test.js
```diff
@@ -37,8 +37,11 @@ describe('History::toJSON', function() {
...
     const history = new History({ maxHistory: Infinity })
 
-    history.append(action).resolve(true)
+    let head = history.append(action).resolve(true)
+    let json = JSON.parse(JSON.stringify(history))
 
-    expect(history).toMatchSnapshot()
+    expect(json.head).toEqual(head.id)
+    expect(json.root).toEqual(head.parent.id)
+    expect(json.size).toEqual(2)
```
mackermedia: Worth adding a test or two to see that you can access `revisions` and possibly that a revision contains `status`, `payload`, and `timestamp` ?

mackermedia:
👍

efatsi:
Nice stuff. Have you considered what it would look like to be able to checkout different revision states within an action? I think the solution you have in place here should be able to support that.

nhunzaker:
> Have you considered what it would look like to be able to checkout different revision states within an action

@efatsi Yes. Actually it has me thinking a lot about how history is stored. We could store every revision in the tree, then insert new revisions as actions are updated and disable old revisions. Most of the bones of that exist now.

It's worth thinking about. I could definitely see a future where you could step through each action iteration.

PR #365, opened 2017-07-11
Add debug mode hook

nhunzaker:
Adds the debug hook we need for the developer tools. This is related to:
https://github.com/vigetlabs/microcosm-devtools/pull/25

screen shot 2017-07-11 at 5 05 47 pm

### DIFF
--- a/src/install-devtools.js
+++ b/src/install-devtools.js
@@ -1,0 +1,9 @@
+export default function installDevtools(repo) {
+  let hook = global.__MICROCOSM_DEVTOOLS_GLOBAL_HOOK__
+
+  if (hook) {
+    hook.emit('init', repo)
+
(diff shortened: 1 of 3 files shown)


efatsi:
so still would require some configuration `{debug: true}`, but that's a pretty low barrier 👍

nhunzaker:
@efatsi Yeah, I figured you would do `{ debug: process.env.NODE_ENV !== 'production' }` once and forget about it.

Basically, I don't want to rely on the `NODE_ENV` environment variable, and I don't want this to automatically run in production. What do you think?

prayasht:
Awesome! This will be super handy. I like the idea of it being in debug mode by default and having to set the env variable to production once it's ready to go.

Issue #367, opened 2017-07-13
Returning an Action Param that is a Function (but not a thunk)


solomonhawk:
I'm not 100% sure whether this would be considered a bug or feature request - so I tagged it as both and as a question. Changing this _may_ be considered a breaking change but I wanted to get some input on that, too.

**What kind of issue is this?** (check at least one)

- [x] Report a bug
- [x] Feature Request
- [x] Question
- [ ] Discussion
- [ ] Documentation
- [ ] Other, please describe:

**What version of Microcosm is this for**

All, (using 12.9.0-beta3 now)

**Steps to Reproduce/Example** (if applicable)

Say you have an action:

```js
function makeAThing(doSideEffect) {
  return doSideEffect
}
```

And you call this action with a function as the argument:

```js
repo.push(makeAThing, function() {
  launchSatelliteIntoOrbit()
  fireZeMissiles()
  return 'Nothing happened.'
})
```

Microcosm's `coroutine` will treat this as a thunk (function body form) and will call the function `doSideEffect` and then return the action.

**Question**: Should coroutine be able to distinguish between an action that returns a function-param and a true thunk?

**Expected**: If the function-argument has side-effects and the user intends to call it later then it is unexpected for that function to be executed as part of the action resolution step.

**Context**: I'm working on expanding the `react-router` examples with some other use cases. One of those is blocking `history` from a route change and prompting the user for confirmation. `history.block` has two forms, one of which is a callback function that returns `Maybe<String>` - if it returns a string then `history` prompts the user with that string otherwise the route change proceeds. One thing I'm exploring is passing that function straight through as the action argument hence this issue.

nhunzaker:
In 12.12.0

PR #368, opened 2017-07-13
#367 Coroutine can distinguish between thunks and pass-through params

solomonhawk:
See #367 

**Question**: Should we do this? Does this make anything harder to do? Theoretically this could be a breaking change if someone is returning an argument and expecting it to be executed as a thunk (passing a thunk action body as an argument from the `repo.push` callsite) which seems like a potentially legitimate use case.

- [x] Bugfix
- [x] Feature
- Code style update
- Refactor
- Build tools
- Documentation
- Other, please describe:

- [x] Yes (maybe?)
- No

If you were passing a function as an action parameter when calling `repo.push` or `send` and then returning it as the value from within the action body _and_ expecting that function-argument to be evaluated as a thunk as part of Microcosm's action resolution process, then you would need to wrap it in an anonymous function before returning it in the action body. e.g.

```js
// before
function maybeThunk(thunkFn, shouldThunk) {
  if (shouldThunk) {
    return thunkFn // I expect this to be called with (action, repo)
  }
}

// after
function maybeThunk(thunkFn, shouldThunk) {
  if (shouldThunk) {
    return (action, repo) => thunkFn(action, repo)
  }
}
```

When the action body's value is a function it is evaluated as a thunk _unless_ it's identity matches one of the passed-in action params. Wrapping the action in a lambda and passing the arguments through will circumvent this equality check and cause the thunk to be evaluated when microcosm rolls forward processing actions.

- [x] All tests are passing
- [x] `yarn pretty` has been run

**Additional Information**

See #367

### DIFF
--- a/src/coroutine.js
+++ b/src/coroutine.js
@@ -94,7 +94,7 @@ export default function coroutine(
   }
 
   /**
-   * Check for thunks. An escape hatch to direction work with an
+   * Check for thunks. An escape hatch to directly work with an
    * action. It is triggered by returning a function from a
    * command. This middleware will execute that function with the
    * action as the first argument. If the returned function is equal
--- a/src/coroutine.js
+++ b/src/coroutine.js
@@ -97,9 +97,11 @@ export default function coroutine(
    * Check for thunks. An escape hatch to direction work with an
    * action. It is triggered by returning a function from a
    * command. This middleware will execute that function with the
-   * action as the first argument.
+   * action as the first argument. If the returned function is equal
+   * to one of the action's params, assume it's a value to be returned
+   * instead of a thunk.
    */
-  if (isFunction(body)) {
+  if (isFunction(body) && !params.some(param => param === body)) {
     body(action, repo)
 
     return action
--- a/test/unit/middleware/thunk-middleware.test.js
+++ b/test/unit/middleware/thunk-middleware.test.js
@@ -8,4 +8,15 @@ describe('Thunk middleware', function() {
 
     expect(spy).toHaveBeenCalledWith(action, repo)
   })
+
+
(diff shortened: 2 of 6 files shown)


src/coroutine.js
```diff
@@ -99,9 +99,11 @@ export default function coroutine(
...
-   * action as the first argument.
+   * action as the first argument. If the returned function is equal
+   * to one of the action's params, assume it's a value to be returned
+   * instead of a thunk.
    */
-  if (isFunction(body)) {
+  if (isFunction(body) && !params.some(param => param === body)) {
     body(action, repo)
 
     return action
```
nhunzaker: This is great, :+1:. Could you add a unit test for this? Check here:

https://github.com/vigetlabs/microcosm/blob/master/test/unit/middleware/thunk-middleware.test.js

nhunzaker:
_Subtle prod_. Anything I can do here to help out?

solomonhawk:
Just need to make time - tomorrow morning I'll update this and the other PR and then get your approval!

solomonhawk:
@nhunzaker Ready for another review. Updated the changelog, docs, and added a test.

test/unit/middleware/thunk-middleware.test.js
```diff
@@ -8,4 +8,15 @@ describe('Thunk middleware', function() {
...
+
+
+  it('does not treat function action arguments as thunks when they are directly returned', function() {
+    let action = fn => fn
+    let spy = jest.fn()
+    let repo = new Microcosm()
+
+    repo.push(action, spy).onDone(result => {
+      expect(result).toEqual(spy)
+    })
```
nhunzaker: In case we ever make actions async, could you add an `expect.assertions(1)` to the top of this test?

PR #369, opened 2017-07-13
Subsequent domains handlers use updates from prior handlers

nhunzaker:
This commit prevents a bug introduced in the Flow rewrite where we
assigned the current working state when dispatching domain handlers
for the same key. Subsequent domain handlers would not receive state
updates from previous handlers, causing prior work to be "undone".

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "microcosm",
-  "version": "12.9.0-beta3",
+  "version": "12.9.0-beta4",
   "private": true,
   "description": "Flux with actions at center stage. Write optimistic updates, cancel requests, and track changes with ease.",
   "main": "microcosm.js",


src/domain-engine.js
```diff
@@ -132,10 +132,10 @@ class DomainEngine {
     for (var i = 0, len = handlers.length; i < len; i++) {
       var { key, domain, handler } = handlers[i]
 
-      var last = get(state, key)
+      var last = get(result, key)
       var next = handler.call(domain, last, action.payload)
 
-      result = set(state, key, next)
+      result = set(result, key, next)
```
nhunzaker: This is the bug. I added `experimental.const_params` to the flow config to prevent assigning over arguments. However I didn't update these internal assignments. The result is that, with multiple domain handlers operating on the same state key, prior operations would be discarded.

💫

mackermedia:
👍

PR #370, opened 2017-07-17
Rework sanitize process, simplify domain engine

nhunzaker:
This commit reworks the sanitization process to avoid some extra inefficiency. Additionally it cuts some code from the domain engine.

This was bugging me for a while since @efatsi pointed it out. Hopefully this makes things a bit nicer.

### DIFF
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -2,6 +2,8 @@
  * @flow
  */
 
+import { mergeSame } from './utils'
+
 import type Action from './action'
 import type Microcosm from './microcosm'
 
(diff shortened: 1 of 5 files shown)


src/lifecycle.js
```diff
@@ -17,12 +19,9 @@ function sandbox(data: Object, deserialize: boolean) {
         throw error
       }
     }
-
     // Strip out keys not managed by this repo. This prevents children from
     // accidentally having their keys reset by parents.
-    let sanitary = repo.domains.sanitize(payload)
-
-    action.resolve(sanitary)
+    action.resolve(mergeSame(repo.state, payload))
```
nhunzaker: We still need this for patch:

```
      const parent = new Microcosm()
      const child = parent.fork()

      parent.addDomain('top', {
        getInitialState() {
          return false
        }
      })

      child.addDomain('bottom', {
        getInitialState() {
          return false
        }
      })

      parent.patch({ bottom: true })

      expect(parent).not.toHaveState('bottom')
      expect(child).toHaveState('bottom', false)
```

efatsi:
Nice 👍

PR #371, opened 2017-07-21
History is an iterator

nhunzaker:
This commit implements the iterator interface for history. This is important for interoperability with other libraries: specifically Observables.

```
// Enumerate over history
for (var action of repo.history) {
  console.log(action)
}

// Or quickly build an array from history
let activeBranch = Array.from(repo.history)

// Or wait for all history to finish
Observable.of(repo.history).subscribe({ next, error, complete })
```

### DIFF
--- a/src/history.js
+++ b/src/history.js
@@ -10,7 +10,7 @@ import Action from './action'
 import Emitter from './emitter'
 import defaultUpdateStrategy from './default-update-strategy'
 import { merge } from './utils'
-import { BIRTH, START } from './lifecycle'
+import { BIRTH, START, ADD_DOMAIN } from './lifecycle'
 import { type Updater } from './default-update-strategy'
 import { iteratorTag } from './symbols'
 
@@ -118,9 +118,9 @@ class History extends Emitter {
    */
   // $FlowFixMe
   [iteratorTag](): Iterator<Action, void> {
-    var cursor = this.root
+    let cursor = this.root
 
-    return {
+    let iterator = {
       next: () => {
         var next = cursor
 
@@ -130,9 +130,17 @@ class History extends Emitter {
 
         cursor = next == this.head ? null : cursor.next
 
+        // Ignore certain lifecycle actions that are only for
+        // internal purposes
+        if (next && (next.command === BIRTH || next.command === START || next.command === ADD_DOMAIN)) {
+          return iterator.next()
+        }
+
         return { value: next, done: false }
       }
     }
+
+    return iterator
   }
 
   map(fn: (action: Action, index: number) => *, scope?: Object) {
@@ -385,6 +393,10 @@ class History extends Emitter {
     }
 
     return this.head
+  }
+
+  toString() {
+    return this.toArray().join(', ')
   }
 
   /**
--- a/src/history.js
+++ b/src/history.js
@@ -12,6 +12,7 @@ import defaultUpdateStrategy from './default-update-strategy'
 import { merge } from './utils'
 import { BIRTH, START } from './lifecycle'
 import { type Updater } from './default-update-strategy'
+import { iteratorTag } from './symbols'
 
 type HistoryOptions = {
   maxHistory?: number,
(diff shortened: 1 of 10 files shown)


test/unit/history/map.test.js
```diff
@@ -0,0 +1,17 @@
...
+  it('provides the index', function() {
+    const history = new History({ maxHistory: Infinity })
+
+    history.append(action, 'resolve')
+    history.append(action, 'resolve')
+    history.append(action, 'resolve')
+
+    let result = history.map((action, i) => i)
+
+    expect(result).toEqual([0, 1, 2, 3])
```
efatsi: 4 elements in the expected array here because of the root node?

test/unit/history/iterator.test.js
```diff
@@ -0,0 +1,64 @@
...
+    let two = history.append('two', 'resolve')
+    let complete = jest.fn()
+
+    Observable.of(history).subscribe({ complete })
+
+    one.resolve()
+    two.resolve()
+
+    expect(complete).toHaveBeenCalled()
+  })
```
efatsi: think it's valuable to test the error cases here?

test/unit/history/map.test.js
nhunzaker: Yeah. This gets back to another discussion we never finished: should we ignore internal actions like `birth` and `start`?

nhunzaker: I vote yes.

src/utils.js
```diff
@@ -219,3 +218,10 @@ export function update(
 
   return set(state, path, next)
 }
+
+/**
+ * A couple of methods use identity functions. This avoids duplication.
+ */
+export function identity<T>(n: T): T {
```
nhunzaker: It's a generic. This is saying "This function will return the same type as it was given"

test/unit/history/map.test.js
efatsi: second

src/utils.js
nhunzaker: `<T>` is what Flow uses in their examples, but this could be clearer. Any thoughts? maybe `<Type>`?

nhunzaker: Or maybe I could just get away with `identity (n: *) { return n }`. Flow might infer that fine.

src/history.js
```diff
@@ -115,16 +116,36 @@ class History extends Emitter {
...
-      if (cursor == this.head) break
-      cursor = cursor.next
+        if (next == null) {
+          return { done: true }
+        }
+
+        cursor = next == this.head ? null : cursor.next
+
+        return { value: next, done: false }
+      }
```
efatsi: conceptually, i think swapping `next` with `cursor` and `cursor` with `next` makes more sense to me, but ... I think this is close to as clear as possible

efatsi: except for what `[iteratorTag]()` means - but i'll chalk that up to my typical lack of JS chops, and the fact that i have no idea what `const $Symbol = isFunction(Symbol) ? Symbol : {}` would do

nhunzaker: 👍

test/unit/history/iterator.test.js
nhunzaker: Yes.

src/history.js
nhunzaker: We need to provide a polyfil for [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol), which is used to describe special properties that indicate that an object provides a particular interface (like enumerability)

nhunzaker: Here's a good overview:
https://babeljs.io/learn-es2015/#ecmascript-2015-features-iterators-for-of

efatsi: is the `iteratorTag` a workaround in lieu of bringing in the polyfill?

nhunzaker: Yeah. `'@@iterator'` is what library authors seem to be falling back to.

Issue #372, opened 2017-07-24
A Presenter's "update" method is not run when the model changes


djmccormick:
**What kind of issue is this?** (check at least one)

- [x] Report a bug
- [x] Feature Request
- [x] Question
- [ ] Discussion
- [ ] Documentation
- [ ] Other, please describe:

**What version of Microcosm is this for**

I ran the example code in `12.9.0-beta4` and `12.8.0` with the same results.

**Steps to Reproduce/Example** (if applicable)

I expected that a Presenter's `update` method to be run whenever the Presenter's model changes, but this doesn't appear to be the case. [The documentation](https://github.com/vigetlabs/microcosm/blob/master/docs/api/presenter.md#updaterepo-nextprops-nextstate) states:

> update is always executed after the latest model has been calculated.

Here is a a bit of example code illustrating the behavior I'm experiencing:

	import React from 'react';
	import ReactDOM from 'react-dom';

	import Microcosm from 'microcosm';
	import Presenter from 'microcosm/addons/presenter'

	//
	// SETUP THE TEST
	//

	// Action
	function increment() {
		return true;
	}

	// Domain
	class CounterDomain {
		getInitialState() {
			return 0;
		}

		register() {
			return {
				[increment]: this.handleIncrement
			};
		}

		handleIncrement(state) {
			return state + 1;
		}
	}

	// Presenter
	class Test extends Presenter {
		setup(repo) {
			repo.addDomain('counter', CounterDomain);
		}

		getModel() {
			return {
				counter: state => state.counter
			};
		}

		ready(repo, props, state) {
			console.log('ready', this.model.counter);
		}

		update(repo, nextProps, nextState) {
			console.log('update', this.model.counter);
		}

		render() {
			console.log('render', this.model.counter);

			return (
				<div>
					Counter is at: {this.model.counter}
					<hr/>
					<a href="" onClick={this.handleClick.bind(this)}>Increment</a>
				</div>
			);
		}

		handleClick(evt) {
			evt.preventDefault();
			this.repo.push(increment);
		}
	}

	//
	// RENDER THE TEST
	//

	ReactDOM.render(
		<Presenter repo={new Microcosm()}><Test/></Presenter>,
		document.getElementById('root')
	);

Here's a screenshot after clicking the "Increment" link six times (notice we never see an "update" string logged):

[...]

PR #375, opened 2017-07-24
Improve specificity of Presenter `update` documentation

solomonhawk:

- Bugfix
- Feature
- Code style update
- Refactor
- Build tools
- [x] Documentation
- Other, please describe:

- Yes
- [x] No

- [x] All tests are passing
- [x] `yarn pretty` has been run

**Additional Information**

Includes more specific/accurate information about how/when `update` is called within Presenters. Specifically tries to disambiguate expectations similar to #372.

### DIFF
--- a/docs/api/presenter.md
+++ b/docs/api/presenter.md
@@ -198,7 +198,7 @@ class PlanetsList extends Presenter {
 
 ### `update(repo, nextProps, nextState)`
 
-Called when a presenter gets new props. This is useful for secondary
+Called when a presenter _gets new props_. This is useful for secondary
 data fetching and other work that must happen when a Presenter receives
 new information.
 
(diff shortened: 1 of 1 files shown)


docs/api/presenter.md
```diff
@@ -222,7 +222,9 @@ class Planet extends Presenter {
 }
 ```
 
-`update` is always executed after the latest model has been calculated.
+In order for this hook to be useful, we ensure that `update` is executed only after the latest model has been calculated.
+
+**NOTE:** `update` is not necessarily always called when the model changes! It only gets called when the `props` sent to Presenter change or when `state` changes within the Presenter.
```
nhunzaker: Do you mind omitting the `necessarily` here?

nhunzaker:
Thanks!

PR #377, opened 2017-07-26
Cache domain handlers when possible

nhunzaker:
This commit changes the way we store snapshots such that they have context into their last operation. This lets us memoize domain handlers. Basically:

1. Is the prior snapshot value the same?
2. Is the prior action status the same?
3. Is the prior action payload the same?

If so, don't invoke the domain handler again!

Testing this in the wild, this has cut extra domain handler invocation for lots of async tasks by about 30%. Pretty neat.

I also believe that we could use this new approach for reporting in the developer tools. We _could_ eventually store each of these snapshots as a diff.

Additionally, if we were to support sub-domains #184, this caching would become more and more powerful as your key paths became more specific. Breaking up complicated state management would get faster.

### DIFF
--- a/src/domain-engine.js
+++ b/src/domain-engine.js
@@ -85,7 +85,7 @@ class DomainEngine {
     return domain
   }
 
-  dispatch(action: Action, state: Object, snapshot: Object) {
+  dispatch(action: Action, state: Object, snapshot: Snapshot) {
     let handlers = this.register(action)
     let result = state
 
--- a/src/domain-engine.js
+++ b/src/domain-engine.js
@@ -92,21 +92,21 @@ class DomainEngine {
     for (var i = 0, len = handlers.length; i < len; i++) {
       var { key, source, handler } = handlers[i]
 
-      var last = get(result, key)
-      var current = get(snapshot.last, key)
+      var base = get(result, key)
+      var head = get(snapshot.last, key)
 
       if (
-        // If the state different
-        last !== current ||
+        // If the reference to the prior state changed
+        base !== head ||
         // Or the payload is different
         action.payload !== snapshot.payload ||
         // or the status is different
         action.status !== snapshot.status
       ) {
-        var next = handler.call(source, last, action.payload)
-
-        result = set(result, key, next)
+        // Yes: recalculate state from the base
+        result = set(result, key, handler.call(source, base, action.payload))
       } else {
+        // No: use the existing snapshot value (memoizing the domain handler)
         result = set(result, key, get(snapshot.next, key))
       }
     }
--- a/src/domain-engine.js
+++ b/src/domain-engine.js
@@ -85,7 +85,7 @@ class DomainEngine {
     return domain
   }
 
-  dispatch(state: Object, action: Action): Object {
+  dispatch(action: Action, state: Object, snapshot: Object) {
     let handlers = this.register(action)
     let result = state
 
@@ -93,9 +93,22 @@ class DomainEngine {
       var { key, source, handler } = handlers[i]
 
       var last = get(result, key)
-      var next = handler.call(source, last, action.payload)
+      var current = get(snapshot.last, key)
 
-      result = set(result, key, next)
+      if (
+        // If the state different
+        last !== current ||
+        // Or the payload is different
+        action.payload !== snapshot.payload ||
+        // or the status is different
+        action.status !== snapshot.status
+      ) {
+        var next = handler.call(source, last, action.payload)
+
+        result = set(result, key, next)
+      } else {
+        result = set(result, key, get(snapshot.next, key))
+      }
     }
 
     return result
--- a/src/history.js
+++ b/src/history.js
@@ -278,15 +278,14 @@ class History extends Emitter {
     console.assert(action, 'History should never reconcile ' + typeof action)
 
     let focus = action
-
     while (focus) {
       this._emit('update', focus)
 
       if (focus === this.head) {
         break
-      } else {
-        focus = focus.next
       }
+
+      focus = focus.next
     }
 
     this.archive()
--- a/src/history.js
+++ b/src/history.js
@@ -279,7 +279,9 @@ class History extends Emitter {
 
     let focus = action
     while (focus) {
-      this._emit('update', focus)
+      if (focus.command !== START) {
+        this._emit('update', focus)
+      }
 
       if (focus === this.head) {
         break
--- a/src/microcosm.js
+++ b/src/microcosm.js
@@ -206,18 +206,12 @@ class Microcosm extends Emitter implements Domain {
    * that, when rolling back to this action, it always has a state value.
    */
   createSnapshot(action: Action) {
-    let state = this.recall(action.parent)
-
-    let snapshot: Snapshot = {
-      last: state,
-      next: state,
+    this.snapshots[action.id] = {
+      last: this.state,
+      next: this.state,
       status: action.status,
       payload: action.payload
     }
-
-    this.snapshots[action.id] = snapshot
-
-    return snapshot
   }
 
   /**
@@ -226,7 +220,7 @@ class Microcosm extends Emitter implements Domain {
   updateSnapshot(action: Action) {
     // Fall back to creating a snapshot if it does not exist in the event
     // an action is in progress while a fork is being created
-    let snap = this.snapshots[action.id] || this.createSnapshot(action)
+    let snap = this.snapshots[action.id]
     let last = this.recall(action.parent)
 
     if (this.parent) {
--- a/src/microcosm.js
+++ b/src/microcosm.js
@@ -223,7 +223,7 @@ class Microcosm extends Emitter implements Domain {
    */
   createSnapshot(action: Action): Snapshot {
     let snapshot: Snapshot = {
-      last: this.rebase(action),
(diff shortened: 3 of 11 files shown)


src/microcosm.js
```diff
@@ -206,24 +206,44 @@ class Microcosm extends Emitter implements Domain {
...
 
   /**
    * Update the state snapshot for a given action
    */
   updateSnapshot(action: Action) {
-    let next = this.recall(action.parent)
+    // Fall back to creating a snapshot if it does not exist in the event
+    // an action is in progress while a fork is being created
+    let snap = this.snapshots[action.id] || this.createSnapshot(action)
+    let last = this.recall(action.parent)
```
efatsi: invalid to go with `let last = snap.last`?

src/domain-engine.js
```diff
@@ -85,17 +85,30 @@ class DomainEngine {
     return domain
   }
 
-  dispatch(state: Object, action: Action): Object {
+  dispatch(action: Action, state: Object, snapshot: Object) {
```
prayasht: This is cool Nate. I'm assuming the `dispatch` method is what actually sends actions (and their payloads) to the store. What exactly does the handler do here?

src/microcosm.js
nhunzaker: Oh, good idea. Let me see.

nhunzaker: It appears that we can not. Though I'd like to figure out how to explain why.

src/domain-engine.js
nhunzaker: Yeah. Exactly. The handlers are cached prior invocations of `register()` on domains so that we don't have to call them every time we dispatch an action.

The handlers are what actually modify state, like the value part of register:

```javascript
const Domain = {
  register() {
    return {
      [action]: handler
    }
  }
}
```

src/microcosm.js
nhunzaker: Ahh. Yes. We can do this. It looks like the `START` action that kicks things off is never getting assigned to a snapshot state.

That's a bummer. Adding START allowed me to avoid some conditional logic, but now it's causing problems. Maybe I can figure out how to remove it.

nhunzaker: Okay. So this optimization actually depends on `priorSnapshot.next` and `thisSnapshot.last` to being out of sync. I don't like that this is kind of an obtuse, round-about, linked list. I'm going to see if I can fix that.

nhunzaker: I think I have it.

The `last` reference in the snapshot refers to the last state that the revision was calculated with. When we can compare that reference to the current working state to say "did my assumptions change? if so, recalculate this domain handler"

Basically, `last` sort of works as a change diffing tool.

src/domain-engine.js
```diff
@@ -85,17 +85,30 @@ class DomainEngine {
...
+        action.payload !== snapshot.payload ||
+        // or the status is different
+        action.status !== snapshot.status
+      ) {
+        // Yes: recalculate state from the base
+        result = set(result, key, handler.call(source, base, action.payload))
+      } else {
+        // No: use the existing snapshot value (memoizing the domain handler)
+        result = set(result, key, get(snapshot.next, key))
+      }
```
efatsi: this is super clean

src/microcosm.js
```diff
@@ -206,24 +206,38 @@ class Microcosm extends Emitter implements Domain {
...
+    }
   }
 
   /**
    * Update the state snapshot for a given action
    */
   updateSnapshot(action: Action) {
-    let next = this.recall(action.parent)
+    // Fall back to creating a snapshot if it does not exist in the event
+    // an action is in progress while a fork is being created
```
efatsi: this comment doesn't seem relevant to the code here

src/microcosm.js
```diff
@@ -206,24 +206,38 @@ class Microcosm extends Emitter implements Domain {
...
+    let snap = this.snapshots[action.id]
+    let last = this.recall(action.parent)
 
     if (this.parent) {
-      next = merge(next, this.parent.recall(action))
+      last = merge(last, this.parent.recall(action))
     }
```
efatsi:
```
let last = this.recall(action.parent)
if (this.parent) {
  last = merge(last, this.parent.recall(action))
}
```
should this functionality of reaching up to the parent be moved into the `recall` method? or maybe a `deepRecall` or something. i'm imagining what happens if the parent has a parent.

efatsi:
really cool stuff, the tests are a bit dense and tough to follow since you're using the same function to define an action as well as what the domain calls in response to the action, but overall 👍

test/integration/reexecuting-domain-handlers.test.js
```diff
@@ -0,0 +1,180 @@
...
+    one.resolve(1)
+
+    expect(addOne).toHaveBeenCalledTimes(1)
+
+    expect(addTwo).toHaveBeenCalledTimes(1)
+
+    expect(repo.state).toEqual({ counterOne: 1, counterTwo: 2 })
+  })
+
+  it('does not call the same domain handler twice when two domains listen to the same action from a different key', function() {
```
efatsi: this test in particular i'm having a tough time following, the "same action from a different key" doesn't line up with what i see below, also why each action is called 3 times and not less is a mystery to me

src/microcosm.js
nhunzaker: Thank you. A relic from a prior update.

src/microcosm.js
nhunzaker: Excellent question.

Right now, parents always update their snapshots before their children because they are created first, so they subscribe to history first. This means they get executed first on the list of callbacks whenever history emits update.

So we don't have to do this. But I want to give it a shot. I think it might avoid some edge cases in the future if we start exploring that.

test/integration/efficiency.test.js
```diff
@@ -80,7 +80,7 @@ describe('Efficiency', function() {
     two.resolve()
     one.resolve()
 
-    expect(handler).toHaveBeenCalledTimes(3)
+    expect(handler).toHaveBeenCalledTimes(2)
```
nhunzaker: Fun little bonus :)

efatsi: you did say 30% improvement right ;)?

nhunzaker:
@efatsi I've made a new `rebase` method that performs the operations you mentioned regarding `recall`. I've also taken a second pass to improve the clarity of those tests.

src/microcosm.js
```diff
@@ -195,35 +195,57 @@ class Microcosm extends Emitter implements Domain {
...
+   */
+  rebase(action: Action) {
+    let state = this.recall(action.parent)
+
+    if (this.parent) {
+      state = merge(state, this.parent.recall(action))
+    }
+
+    return state
+  }
```
nhunzaker: Devtools rely on `recall` to get the state snapshot for an action, so I'm afraid to touch it. We should come up with a public API for getting the snapshot for an action, use that in the dev tools, and revisit this.

src/microcosm.js
```diff
@@ -112,7 +112,7 @@ class Microcosm extends Emitter implements Domain {
 
     this.history = this.parent ? this.parent.history : new History(options)
 
-    this.snapshots = {}
+    this.snapshots = Object.create(this.parent ? this.parent.snapshots : null)
```
nhunzaker: This is new. By using prototypal inheritance, forks gain access to prior snapshots without needing to "sync up" as actions update that they couldn't track. Accessing a snapshot prior to when a fork was created causes prototypal access from the parent.

src/microcosm.js
```diff
@@ -193,37 +193,61 @@ class Microcosm extends Emitter implements Domain {
...
 
     if (!action.disabled) {
-      next = this.domains.dispatch(next, action)
+      snap.next = this.domains.dispatch(action, last, snap)
+    } else {
+      snap.next = last
     }
 
-    this.state = this.snapshots[action.id] = next
+    this.snapshots[action.id] = merge(snap, { last, status: action.status, payload: action.payload })
```
nhunzaker: Since we share data, we can't use nested mutation. The overhead here appears to be negligible.

test/integration/snapshots-in-forks.test.js
```diff
@@ -0,0 +1,44 @@
+import Microcosm from '../../src/microcosm'
+
+describe('Snapshots', function() {
+  it('forks accommodate actions that were pushed before the parent was forked', function() {
```
nhunzaker: This test specifically covers an edge case I found where:

1. You create a presenter
2. That presenter pushes an action in the setup method
3. A react router route embeds another presenter as a child

This put Microcosm into a state where the fork form the child presenter did not have a snapshot for the parent. The prototypal inheritance update addresses this problem.

src/history.js
```diff
@@ -278,15 +278,16 @@ class History extends Emitter {
     console.assert(action, 'History should never reconcile ' + typeof action)
 
     let focus = action
-
     while (focus) {
-      this._emit('update', focus)
+      if (focus.command !== START) {
```
nhunzaker: I think a good next step for History is to figure out how to remove START, BIRTH, and ADD_DOMAIN

src/microcosm.js
prayasht: That is badass

PR #385, opened 2017-08-07
Allow a cascade of options. Add static defaults to domains and effects

### LINKED ISSUE #384: Default options for effects and domains
On Microcosm, we have the ability to set default options:

```
class Repo extends Microcosm {
  static defaults {
    debug: true,
    maxHistory: 20
  }
}
```

We should do this for domains and effects. Additionally -- what if we allowed the configuration to cascade?

nhunzaker:
Domains and effects can have a `defaults` class property to provide default setup options:

```javascript
class Counter {
  static defaults = {
    start: 0
  }

  setup (repo, { start }) {
    console.log(start) // 0
  }
}

let repo = new Microcosm()

repo.addDomain('counter', Counter) // default start is 0
```

---

Fixes #384.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 # Edge
 
 - Microcosm ships with ES6 and UMD bundles
+- Domains and Effects can implement a `defaults` static object to
+  provide default setup options.
 
 # 12.9.0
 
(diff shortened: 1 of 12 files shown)


efatsi:
nice stuff 👍

PR #386, opened 2017-08-08
Do not return undefined when getting null with no fallback

nhunzaker:
Just got hit by this nasty bug. Getting `null` out of an object was returning undefined because the loose null check returned an undefined fallback:

```
get({ prop: null }, 'prop') // undefined, but should be null
```

This is because we trigger the fallback on null. The problem is that a value is present here, so we should respect it when there is no fallback.


mackermedia:
👍  Niiiiiice

nhunzaker:
Dang. Looks like this caught a bug in change-tree switching between undefined and null.

Issue #380, opened 2017-08-01
Only test the public API


nhunzaker:
I want to start experimenting with tools to reduce the build size, like ([buble](buble.surge.sh) or [Google Closure Compiler](https://github.com/google/closure-compiler-js)), and improve execution time (like [prepack](https://github.com/facebook/prepack)).

The problem is that, for the sake of test coverage, we directly test private APIs quite a bit. Frankly, I'm really not sure if this is a good or bad thing. However it's definitely a bad thing for tools that change code during the build process. 

Ideally, our test suite always runs against the code that hits production.

- [ ] action
- [ ] compare-tree
- [ ] domain
- [ ] effect
- [ ] history
- [ ] key-path
- [ ] microcosm
- [ ] middleware
- [ ] utils

PR #387, opened 2017-08-08
Test registration in terms of public API

nhunzaker:
Related to #380. I'm chipping away at only using the public API for testing. That way, we can eventually do some advanced build optimizations and still ensure the correctness of Microcosm.


src/history.js
```diff
@@ -214,9 +215,13 @@ class History extends Emitter {
...
     this.size += 1
 
+    action.on('change', this.reconcile, this)
+
     this._emit('append', action)
 
-    action.on('change', this.reconcile, this)
+    if (status && action.command !== START) {
+      this.reconcile(action)
+    }
```
nhunzaker: Fun little bug. if you did `repo.append(action, 'status')`, reconciliation wouldn't happen the first time!

efatsi: nice little catch

test/unit/microcosm/lifecycle.test.js
```diff
@@ -1,19 +1,48 @@
...
+
+    let repo = new Repo()
+    let spy = jest.spyOn(repo, 'getInitialState')
+
+    repo.getInitialState()
 
-    expect(test).toThrow()
+    expect(spy).toHaveBeenCalledTimes(1)
   })
 })
```
efatsi: have these methods been called infinitely at times? curious what these tests accomplish

test/unit/registration.test.js
```diff
@@ -0,0 +1,123 @@
+import Microcosm from '../../src/microcosm'
+import { ALIASES } from '../../src/get-registration'
```
efatsi: thoughts on renaming ALIASES to STATUSES? or would that possibly be a breaking change

test/unit/registration.test.js
```diff
@@ -0,0 +1,123 @@
...
+            }
+          }
+        }
+
+        let repo = new Repo()
+
+        repo.append(action, status)
+
+        expect(handler).toHaveBeenCalled()
+      })
```
efatsi: i believe this test and the one before it have identical test bodies

test/unit/registration.test.js
```diff
@@ -0,0 +1,123 @@
...
+          }
+        }
+      }
+
+      let repo = new Repo()
+
+      expect(repo.prepare(getUser)).toThrow(
+        'getUser key within a registration is undefined. Is it being referenced correctly?'
+      )
+    })
```
efatsi: `'uses the command name when it can'` doesn't seem to line up with the fact that this test is testing invalid registration

test/unit/registration.test.js
```diff
@@ -0,0 +1,123 @@
...
+      }
+
+      let repo = new Repo()
+      let spy = jest.spyOn(repo, 'deserialize')
+
+      repo.deserialize()
+
+      expect(spy).toHaveBeenCalledTimes(1)
+    })
+  })
```
efatsi: you're already testing this in the lifecycle tests

test/unit/microcosm/lifecycle.test.js
nhunzaker: Yes. This was a case when we first added `register` to Microcosm. I can't see how it would happen though. This can go.

nhunzaker: Further more it should be covered by deserialize/serialize tests.

test/unit/registration.test.js
nhunzaker: Naw. It's not a breaking change. They are only used internally.

test/unit/registration.test.js
nhunzaker: One tests the canonical status, the other the alias. There was a point where you'd call `action.resolve()` and register to `[action.done]`. We've updated this so that the canonical way is to reference `action.resolve` in both cases. But to avoid a breaking change, we support both `resolve` and `done`

Similarly with `reject` and `error`.

test/unit/registration.test.js
nhunzaker: Good catch.

nhunzaker: I think this meant to say "It includes the command name in the error"

test/unit/registration.test.js
nhunzaker: Yeah. Bugger. Killing it.

PR #388, opened 2017-08-09
More public tests

nhunzaker:
Adds remaining testing in terms of a public API.

Ooff. I caught some things along the way.


package.json
```diff
@@ -32,6 +32,7 @@
   "license": "MIT",
   "jest": {
     "testURL": "http://localhost:3000",
+    "testEnvironment": "node",
```
nhunzaker: Test in the node environment to catch cases where we might depend on `window` without a fallback for node (using `global`)

src/history.js
```diff
@@ -369,7 +371,7 @@ class History extends Emitter {
    * Set the limit of the history object.
    */
   setLimit(limit: number) {
-    this.limit = Math.max(0, limit)
+    this.limit = Math.max(DEFAULTS.maxHistory, limit)
```
nhunzaker: This is a fun one. We always want the minimum limit to be 1 right now, so that there's always a "parent" action to derive a base state snapshot from.

src/history.js
```diff
@@ -343,9 +343,11 @@ class History extends Emitter {
       }
     }
 
-    root.prune()
+    if (root !== this.root) {
+      this.root = root
+      root.prune()
```
nhunzaker: Just a small thing. root.prune() would get called every time, even if the root didn't change.

src/utils.js
```diff
@@ -59,12 +59,12 @@ export function merge(...args: Array<?Object>): Object {
   let subject = null
 
   for (var i = 0, len = args.length; i < len; i++) {
-    copy = copy || args[i]
-
-    if (copy == null) {
+    if (isObject(args[i]) === false) {
       continue
     }
```
nhunzaker: Caught `merge(true, true, {})` would return true. This is because `||` was using the `true` value, and no keys in the object meant that the object casting process would never happen.

test/integration/reconciliation.test.js
```diff
@@ -69,15 +69,16 @@ describe('Reconciliation', function() {
...
+    let two = repo.append('two')
+    let three = repo.append('three')
 
     three.resolve()
     two.resolve()
     one.resolve()
 
-    expect(Object.keys(repo.snapshots)).toEqual([three.id])
+    // Two is in here because it is the  parent of three, which we need as a "base state"
+    expect(Object.keys(repo.snapshots)).toEqual([two.id, three.id])
```
nhunzaker: This happened when I bumped the limit to a minimum of 1. This is correct behavior.

test/unit/action/abandon.test.js
```diff
@@ -1,17 +0,0 @@
...
-
-    parent.adopt(one)
-    parent.adopt(two)
-
-    parent.abandon(one)
-    parent.abandon(one)
-
-    expect(parent.children).toEqual([two])
-  })
-})
```
nhunzaker: Removed. These are covered by other test suites and dig too much into internals.

test/unit/action/autobinding.test.js
```diff
@@ -1,22 +0,0 @@
...
-        callback(true)
-      }
-
-      passByValue(action[state])
-
-      expect(action.status).toEqual(state)
-      expect(action.payload).toEqual(true)
-    })
-  })
-})
```
nhunzaker: Removed. These are covered by other test suites and dig too much into internals.

test/unit/action/constructor.test.js
```diff
@@ -1,18 +0,0 @@
...
-  })
-
-  it('an action can be set to a specific status', function() {
-    let repo = new Microcosm()
-
-    let action = repo.append('test', 'resolve')
-
-    expect(action.is('resolve')).toBe(true)
-  })
-})
```
nhunzaker: Removed. These are covered by other test suites and dig too much into internals.

PR #392, opened 2017-08-11
Make register object optional

nhunzaker:
Makes returning an object optional. Cuts an object allocation on every new action push because we don't need to return an object in the default Microcosm register method.

### DIFF
--- a/src/microcosm.js
+++ b/src/microcosm.js
@@ -394,7 +394,7 @@ class Microcosm extends Emitter implements Domain {
    * The default registration method for Microcosms
    */
   register() {
-    return {}
+    return null
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/microcosm.js
```diff
@@ -394,7 +394,7 @@ class Microcosm extends Emitter implements Domain {
    * The default registration method for Microcosms
    */
   register() {
-    return {}
+    return null
```
prayasht: What's the case for returning `null` here vs. an empty object?

nhunzaker: Abysmally low memory savings, but also avoiding all of the checks to see if the registration supports a specific action.

PR #393, opened 2017-08-11
Run tests against minified build

nhunzaker:
This PR sets up a method on `it` and `describe` to disable tests when running a production build. This allows us to test internals for a few modules while minifying private properties in production.

It also adds a plugin to report the output size of modules:

prod-test


scripts/bundle
```diff
@@ -55,9 +59,12 @@ if (options.minify) {
...
       mangle: {
         toplevel: true,
-        // https://github.com/mishoo/UglifyJS2#mangle-properties-options
         properties: {
-          regex: /^_/
+          regex: /^_/,
+          cache: {
+            cname: -1,
+            props: new Dictionary()
+          }
```
nhunzaker: This is pretty cool, and probably worth blogging about. You can share a dictionary of minified properties between builds. This will use the same minification dictionary across all modules for a build.

src/install-devtools.js
```diff
@@ -1,6 +1,8 @@
+const key = '__MICROCOSM_DEVTOOLS_GLOBAL_HOOK__'
```
nhunzaker: Bug caught by running a prod build! the `__` here was picking up our regex, so dev mode was broken in the minified build.

test/addons/presenter.test.js
```diff
@@ -1352,7 +1352,7 @@ describe('::send', function() {
       </Parent>
     )
 
-    wrapper.find('ActionButton').simulate('click')
+    wrapper.find(ActionButton).simulate('click')
```
nhunzaker: Names get minified, so test against the ActionButton component, not the string name.

PR #400, opened 2017-08-24
Fix strange ref warning

nhunzaker:
I hit a strange warning with refs, and we need to update the way the ref is assigned in ActionForm for React 16 anyway.


src/addons/action-form.js
```diff
@@ -45,7 +51,13 @@ class ActionForm extends React.PureComponent {
   }
 
   submit(event: Event) {
-    let form = this.refs.form
+    let form = this.form
+
+    console.assert(
+      form,
+      'ActionForm submit was called before this component was mounted.'
+    )
```
nhunzaker: Added a basic assertion here for strict mode. This should never happen, but impossible states like to happen in the wild... Hopefully this makes debugging easier.

nhunzaker: Ooh actually this could happen post-mount too.

PR #401, opened 2017-08-25
Multiple domain handlers

nhunzaker:
This commit allows Domains and Effects to provide a list of handlers. 

### Domains Reduce

These handlers reduce over state. This should allow domain handlers to be even simpler operations.

This comes really in handle for sorting behavior:

```javascript
import { sortBy } from 'lodash'

class Widgets {
  getInitialState() {
    return []
  }
  add(items, widget) {
    return items.concat(widget)
  }
  sort(items) {
    return sortBy(items, 'id')
  }
  register () {
    return {
      [createWidget]: [this.add, this.sort]
    }
  }
}
```

### Effects Iterate

A list of handlers given to an effect are called sequentially, however there is no reducing behavior:

```javascript
class MyEffect {
  register() {
    return {
      [createWidget]: [
        (repo, action) => log(action.status),
        (repo, action) => log(action.payload),
      ]
    }
  }
}
```

### Presenters do not chain

`intercept()` only allows one callback. I don't actually know what to do here, should it reduce? If it just iterates, what do we return back from `send`?


mackermedia:
`Presenter.intercept` seems like an okay thing to not reduce/chain. It's saying "hey I want to do something about the event/data here" which can have its own async or nested operations

src/domain-engine.js
```diff
@@ -110,7 +110,8 @@ class DomainEngine {
         action.status !== snapshot.status
       ) {
         // Yes: recalculate state from the base
-        result = set(result, key, handler.call(source, base, action.payload))
+        var next = pipeline(steps, action.payload, base, scope)
```
mackermedia: `pipeline()` function looks like it "acts on the steps" and returns the result. Maybe consider naming it something more explicit? `reduceSteps()`, `executeSteps`, etc. Also, before I looked at the `pipeline()` implementation, `next` as the result variable name had me thinking it only acted on the first step of the array. `result` is already expected/taken, maybe name it `stepsResult` ?

mackermedia:
Neat-o! 👍

efatsi:
👍

src/domain-engine.js
nhunzaker: Renamed `pipeline()` to `reduce()` and cut the `next` variable assignment in 000d18e

PR #402, opened 2017-08-31
Batch presenter model callbacks when possible

nhunzaker:
This commit sets up Presenter model updates to be async when batching is enabled. **This only affects rendering**, the current model is always available.

This is a side-effect of working around an update issue with React Router. React Router 4 doesn't seem to allow PureComponents as route handlers, so Presenters never update :(.

For my current project, this also provides a nice performance improvement:

save


src/addons/frame.js
```diff
@@ -0,0 +1,14 @@
...
+
+const hasFrame = typeof requestAnimationFrame !== 'undefined'
+
+let requestFrame = setTimeout
+let cancelFrame = clearTimeout
+
+/* istanbul ignore if */
+if (hasFrame) {
+  requestFrame = requestAnimationFrame
+  cancelFrame = cancelAnimationFrame
```
mackermedia: slick

src/addons/shallow-diff.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Diffing algorithm for presenters. We need to recalculate the
+ * model of props are state change, but ignore children.
+ */
+
+export default function shallowDiff(a, b) {
```
mackermedia: should this be called `isShallowDiff` or `shallowDifferent`. Or maybe flip the logic and call it `shallowEqual` ?
Having it end in just `diff` makes me think it's going to return a set of items that are different (like lodash `difference`) or a file change diff

nhunzaker: Ah. Good call. This is like "Shallow equal, but not children". Maybe this should be `shouldUpdateModel`

mackermedia:
Nice find and improvement. One small question about naming, otherwise :+1:

src/addons/shallow-diff.js
mackermedia: also, if it's used internally only, should it live in `src/addons/`?

PR #406, opened 2017-09-13
Add the ability to alias actions

### LINKED ISSUE #405: Add a way to alias actions in Domains
I want a better way to alias actions with string names. This would be a breaking change because it would rework the way string actions are treated.

I would like to consider putting actions right next to domains. A domain module would contain both actions and handlers.

```javascript
export function createUser (params) {
  return axios.post('/users', params)
}

class UsersDomain {
  actions = { createUser }

  add(users, user) {
    return users.concat(user)
  }

  register() {
    return {
      createUser: this.add
    }
  }
}

let repo = new Microcosm()

repo.addDomain('users', UsersDomain)

repo.push('createUsers', { name: 'Bill' }) // adds a new user
repo.push('createUser', { name: 'Bill' }) // Error: Unrecognized action "createUser". Did you mean "createUsers"?
```

When a domain is added it will:

1. Validate that no existing aliases exist for the domain
2. Tag the action with the associated key
3. If a string action is pushed and nothing responds, the action will reject with a standard error
4. If no code subscribes to "action.onError", Microcosm will send out a warning

### Things to figure out

1. Can we do this for Effects too?
2. What happens when two domains alias the same action?
3. If we declare actions ahead of time, Microcosm can know about them. This would let us make a "control box" in the debugger for firing off actions.

nhunzaker:
This change allows actions to be aliased to a given string. When the string name is not recognized, Microcosm raises an assertion error in strict mode.

In order to prevent this from being a breaking change, string actions that are pushed into a repo that are unrecognized through aliasing still work. **However I would like to eventually send out a minor release where this is deprecated, and throw an exception in the next major release**.

The benefit of this change is that we can use it to know about available actions and prevent the need to import actions in places like `ActionButtons`, something I know @mackermedia will love :). It also makes it so that you do not need to use computed keys for actions. Domain registration can actually happen without a function, like so:

```javascript
let repo = new Microcosm()

repo.addDomain('planets', {
  getInitialState() {
    return []
  },

  actions: { addPlanet: name => name },

  register: {
    addPlanet: (planets, params) => planets.concat(params)
  }
})
```

I need this for my GraphQL Microcosm work in order to know what Mutation types are valid, and to support some new debugger features. 

Fixes #405


src/microcosm.js
```diff
@@ -329,10 +259,10 @@ class Microcosm extends Emitter implements Domain {
    * repo.push(createPlanet, { name: 'Merkur' })
    * ```
    */
-  push(command: Command, ...params: *): Action {
+  push(command: any, ...params: *): Action {
```
nhunzaker: Flow still doesn't like that tagged actions are functions with properties that stringify to a unique id. This is a workaround.

src/microcosm.js
```diff
@@ -487,6 +423,80 @@ class Microcosm extends Emitter implements Domain {
   parallel(actions: Action[]) {
     return this.append('GROUP').link(actions)
   }
+
+  /* Private ------------------------------------------------------ */
```
nhunzaker: @efatsi I'm going to start moving private methods into a section like this. If we append an underscore to a method name, it'll get minified in the production pass. These tokens are shared across all of the modules, so if we _did_ need to access a private method in something like `Presenter`, it would be safe.

src/tag.js
```diff
@@ -57,5 +57,5 @@ export default function tag(fn: Command | Tagged, name?: string): Tagged {
   // Mark the function as tagged so we only do this once
   fn.__tagged = true
 
-  return cast
+  return fn
```
nhunzaker: Really happy about this. It looks like you can cast a Function type to an Object in flow to assign properties to it:

```
  let cast: Tagged = fn

  // assign props
  cast.open = symbol + '.open'
  // ...

  // Return the original function to make flow happy
  return fn
```

test/integration/reconciliation.test.js
```diff
@@ -66,7 +66,7 @@ describe('Reconciliation', function() {
     expect(repo).toHaveState('count', 6)
   })
 
-  it('archived actions are removed from the archive', function() {
+  it.dev('archived actions are removed from the archive', function() {
```
mackermedia: mean to leave this test as `.dev` ?

mackermedia:
I probably don't have enough context around the codebase to know if this is A+ and I'm still not up to speed on flow so this is a bit hard to review.

That said, I think the concept and tests are :+1:

test/integration/reconciliation.test.js
nhunzaker: Yeah. I should pick a better name. `.dev` means "only run this in the development test command". Microcosm also tests against the production build, but that doesn't have any of the assertions of the development version.

So this lets me test both dev and prod; taking advantage of assertions while also making sure the production build works.

efatsi:
NIce 👍

PR #407, opened 2017-09-18
Move Presenter batching to Model. Use History scheduler in Model.

nhunzaker:
I realized that the Preact Presenter wasn't updated with the batched scheduling update from the React Presenter. This commit moves all of that logic into the Model, adding a few optimizations along the way:

1. Models use the same update scheduler as history
2. Models do not queue an update if nothing has changed, so:
3. The Presenter mediator no longer needs to extend from PureComponent

For this to work, I had to make it so that the scheduler used by nHistory could be cancelled. For this, I borrowed some ideas from https://github.com/yoshuawuyts/on-idle.

`on-idle` does some neat stuff to re-dispatch a job if it completes too fast, which is nice because you create a minimum window to collect changes. I hit some strangeness when I set this up, but I'd like to explore it in the future.

While I was in there, I've also privatized most of the Model API. This also improves minification.


packages/microcosm/src/addons/model.js
```diff
@@ -22,108 +23,140 @@ function invoke(binding: *, repo: Microcosm, scope: Object): * {
...
+        this._bindings.push({ key, callback })
+      }
+    }
 
-    this.repo.on('change', this.compute, this)
+    if (this._bindings.length) {
+      this._repo.on('change', this._compute, this)
+    } else {
+      this._repo.off('change', this._compute, this)
+    }
```
nhunzaker: If we're only using Observables in the model, then don't listen to the Microcosm instance at all. Instead rely on those Observables to "push" in new updates.

packages/microcosm-preact/src/presenter.js
```diff
@@ -139,13 +138,11 @@ class PresenterMediator extends Component {
 
     this.presenter = props.presenter
     this.repo = this.presenter._requestRepo(context.repo)
-    this._lastRevision = -Infinity
```
nhunzaker: No need to track revisions now. When the Presenter gets new properties/state and renders, it will merge any pending model changes and cancel any delayed jobs to update the model.

packages/microcosm-preact/src/presenter.js
```diff
@@ -139,13 +138,11 @@ class PresenterMediator extends Component {
...
 
     this.model = new Model(this.repo, this.presenter)
 
     // The following methods are autobound to protect scope
     this.send = this.send.bind(this)
-    this._scheduleUpdate = this._scheduleUpdate.bind(this)
```
nhunzaker: Similarly, Presenters don't deal with scheduling any more. This is all handled in the Model.

packages/microcosm/src/addons/model.js
nhunzaker: Hmmm. This is going to attach a new listener without cleaning up the old one. Fixing.

packages/microcosm/src/addons/model.js
```diff
@@ -14,116 +15,186 @@ function isCallable(binding: *): boolean {
...
+          nextBindings[key] = callback
+        }
+      } else if (this.value[key] !== callback) {
+        // Otherwise we just have a primitive value. If the value is
+        // different, push the key/value pair into the next patch and
+        // queue a recompute.
+        recompute = true
+        this._patch[key] = callback
+      }
+    }
```
nhunzaker: Man, this is too many use cases! I'd really like to drop the callable interface at some point and just use primitive values or observables.

nhunzaker: There's a pretty tremendous perf boost if we drop callables too, because it means we can stop listening to the repo naively and "push" new values in as needed - ideally batched.

PR #409, opened 2017-09-18
Engage Microcosm in layers

nhunzaker:
Presenters build a lot of forks. This commit makes it so that most of the overhead of Microcosm is avoided in forks if they do not attach domains or effects.

## Before

```
Conducting fork benchmark...

Count    Setup      Push
-------  ---------  ---------
1,000    12.527ms   27.356ms
10,000   72.586ms   106.089ms
50,000   213.31ms   475.951ms
100,000  390.559ms  775.968ms
```

## After

```
Conducting fork benchmark...

Count    Setup      Push
-------  ---------  --------
1,000    6.894ms    2.48ms
10,000   30.983ms   5.023ms
50,000   147.558ms  19.37ms
100,000  163.918ms  38.091ms
```

### DIFF
--- a/packages/microcosm/src/microcosm.js
+++ b/packages/microcosm/src/microcosm.js
@@ -109,34 +109,18 @@ class Microcosm extends Emitter implements Domain {
 
     this.options = merge(DEFAULTS, this.constructor.defaults, preOptions || {})
 
-    this.parent = this.options.parent
-
-    this.initial = this.parent ? this.parent.initial : this.getInitialState()
-    this.state = this.parent ? this.parent.state : this.initial
-
-    this.history = this.parent ? this.parent.history : new History(this.options)
-
-    this.actions = Object.create(this.parent ? this.parent.actions : null)
-    this.snapshots = Object.create(this.parent ? this.parent.snapshots : null)
-    this.domains = new DomainEngine(this)
-    this.effects = new EffectEngine(this)
-    this.changes = new CompareTree(this.state)
     this.active = true
+    this.parent = this.options.parent
+    this.initial = this.parent ? this.parent.initial : this.getInitialState()
+    this.snapshots = Object.create(this.parent ? this.parent.snapshots : null)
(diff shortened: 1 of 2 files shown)


packages/microcosm/src/microcosm.js
```diff
@@ -109,34 +109,18 @@ class Microcosm extends Emitter implements Domain {
...
-    this.effects = new EffectEngine(this)
-    this.changes = new CompareTree(this.state)
-    this.active = true
-
-    // History moves through a set lifecycle. As that lifecycle occurs,
-    // save snapshots of new state:
-
-    // When an action is first created
-    this.history.on('append', this._createSnapshot, this)
+    this.changes = new CompareTree(this.initial)
```
nhunzaker: We still always need a compare tree per fork. Forks mitigate a lot of the value of the compare tree because we can't take advantage of structural sharing between data subscriptions.

Something to think about...

packages/microcosm/src/microcosm.js
```diff
@@ -154,6 +138,10 @@ class Microcosm extends Emitter implements Domain {
     }
   }
 
+  get state(): Object {
+    return this._recall(this.history.head)
+  }
```
nhunzaker: `repo.state` is now a getter. It'll return the latest snapshot for history. This means we never have to make sure parents and children are synced up.

packages/microcosm/src/microcosm.js
```diff
@@ -496,6 +488,34 @@ class Microcosm extends Emitter implements Domain {
...
+
+    // When an action is first created
+    this.history.on('append', this._createSnapshot, this)
+
+    // When an action snapshot needs updating
+    this.history.on('update', this._updateSnapshot, this)
+
+    // When an action snapshot should be removed
+    this.history.on('remove', this._removeSnapshot, this)
+  }
```
nhunzaker: This here (avoiding snapshots for Microcosms that have no domains) is what made things so much faster.

PR #413, opened 2017-09-19
Add basic microcosm-http implementation

nhunzaker:
This commit adds an example and basic implementation of the microcosm-http library. 

It leans on [axios](https://github.com/mzabriskie/axios), which unfortunately uses Promises (I'd like to avoid requiring a polyfil we don't technically need). Still, axios is rock solid and works server/client side.

I've also added an `onNext` callback to Action, ActionButton, and ActionForm. This makes it easier to listen to a variety of action statuses, as demonstrated in the example.

There's a couple of fun things already:

### Upload progress

upload

### Upload failure

error


packages/microcosm-http/src/http.js
```diff
@@ -0,0 +1,63 @@
...
+
+      let request = axios({
+        // https://github.com/mzabriskie/axios#request-config
+        ...options,
+        cancelToken: source.token,
+        onUploadProgress: trackUploads(action),
+        onDownloadProgress: trackDownloads(action)
+      })
+
+      request.then(response => action.resolve(response.data)).catch(error => {
```
nhunzaker: Again it's a bit of a bummer that we need Promises here. I wonder if we could provide lightweight mock.

packages/microcosm-http/webpack.config.js
```diff
@@ -0,0 +1,50 @@
...
+      const upload = multer({ dest: 'example/uploads' })
+
+      app.use(bodyParser.json())
+      app.use(bodyParser.urlencoded({ extended: true }))
+
+      app.post('/files', upload.array('files', 100), function(req, res, next) {
+        res.json(req.body)
+      })
+    }
+  }
```
nhunzaker: Teaches webpack how to handle file uploads.

packages/microcosm/src/emitter.js
```diff
@@ -6,7 +6,7 @@
 
 import { isFunction } from './utils'
 
-export type Callback = (...args: *[]) => *
+export type Callback = (...args: any[]) => any
```
nhunzaker: Making a note to come back to this. We don't do any type checking of callbacks because they can accept anything and we don't handle their return value.

PR #423, opened 2017-09-21
Microcosm graphql cleanup

nhunzaker:
I'm trying to make a general query format for Microcosm. I want that to be informed by the graphql query stuff.

While I was learning GraphQL I created a bit of a mess. In this PR I've cleaned up a lot of query operations, and I've started to explore what the Presenter might look like:

```javascript
class Example extends GraphPresenter {
  model = Model

  renderItem(item) {
    return (
      <li key={item.id}>
        {item.title} by {item.author.name}
      </li>
    )
  }

  render() {
    const { posts } = this.state.model

    return <ul>{posts.map(this.renderItem, this)}</ul>
  }
}
```


packages/microcosm-graphql/src/resolve.js
```diff
@@ -1,88 +1,30 @@
-import { get } from 'microcosm'
```
nhunzaker: It is probably best to look at this file side-by-side.

Issue #424, opened 2017-09-25
microcosm-addons: ActionForm - Pass form as second argument to action state callbacks


nhunzaker:
Right now it's hard to reset a form when an action completes. You have to assign a ref to the form, which is kind of a bummer.

What if instead we passed the `form` in the callbacks?

```javascript
class MyForm extends React.Component {
  onDone(_payload, form) { 
    form.reset()
  }

  render() {
    return (
      <ActionForm action="doSomething" onDone={this.onDone}>
        ...
      </ActionForm>
    )
  }
}
```

PR #431, opened 2017-10-03
Pass form as second argument to action callbacks

nhunzaker:
Right now it's hard to reset a form when an action completes. This commit adds the form as the second argument of these callbacks, making this easier:

```javascript
class MyForm extends React.Component {
  onDone(_payload, form) { 
    form.reset()
  }

  render() {
    return (
      <ActionForm action="doSomething" onDone={this.onDone}>
        ...
      </ActionForm>
    )
  }
}
```

Fixes: https://github.com/vigetlabs/microcosm/issues/424


efatsi:
👍

Issue #425, opened 2017-09-25
microcosm-addons: ActionForm/ActionForm - Unsubscribe to action when component unmounts


nhunzaker:
If you push an action that results in a state change that unmounts an ActionButton or ActionForm, you can end up in a state where callback props like `onDone` or `onError` are called on an unmounted component.

I don't _think_ this should happen.

nhunzaker:
Got it in #432

PR #432, opened 2017-10-04
ActionForm and ActionButton actions should unsubscribe on mount

nhunzaker:
Before this commit, there was an issue where ActionForm and ActionButton callbacks would still fire if the related components unmounted. This resulted in an error where `setState` would call on an unmounted component for most of the use cases for these callbacks.

Basically:

```javascript
const testAction = function() {
  return Promise.resolve(true)
}

class Test extends React.Component {
  state = {
    status: 'inactive'
  }

  onNext = ({ status }) => {
    this.setState({ status })
  }

  render() {
    return <ActionForm action={testAction} onNext={this.onNext} />
  }
}

let form = ReactDOM.render(<Test />, el)

form.submit()

ReactDOM.unmountComponentAtNode(el)

// Action callbacks are still firing. onNext will raise a setState error in React when the action completes.
```

This commit does a couple of things:

1. Adds `Action::execute`: Moves coroutine behavior to the Action prototype
2. Adds `action._origin`. Actions know what repo they were created with. This is necessary for execute, but could be used for other future work.
3. `ActionQueue`: A helper class for entities that must dispatch and track multiple actions.

The result of this work is that ActionForm and ActionButton now "empty" their queue of actions when they unmount, unsubscribing any callbacks they set on those actions.

---

https://github.com/vigetlabs/microcosm/issues/425


packages/microcosm/src/microcosm.js
```diff
@@ -234,10 +233,10 @@ class Microcosm extends Emitter implements Domain {
    */
   append(command: *, status?: Status): Action {
     if (command in this.actions) {
-      return this.history.append(this.actions[command], status)
+      return this.history.append(this.actions[command], status, this)
```
efatsi: could this be used to instantiate instances of microcosm trees that don't process actions outside of their microcosm branch?

nhunzaker: Good question. I don't think so, but it might be a first step towards it.

We need to do some planning for making a microcosm that doesn't listen to _certain_ actions but not others. Right now every Presenter forks it's associated repo. I'm afraid that it would be hard to tell why an action isn't getting responded to.

This might need to happen inside of a domain. I wonder if there's a concept that we're missing like a public/private domain. Private domains could be setup such that they didn't respond to actions on their parents or siblings. 

It's a good thing for us to sort out for 13.x

PR #434, opened 2017-10-10
Remove compare tree

nhunzaker:
**What: Removes the CompareTree data binding tool**

The CompareTree class was an addition added in the 12.x release of Microcosm that made it easy to subscribe to fine-grained pathways within a Microcosm state tree:

```javascript
let repo = new Microcosm()
// ...setup

repo.on('change:planets,physics.quantumThreshold', (planets, threshold) => {
  // Do something with planets and quantum threshold (what is that even?)
})
```

It does this extremely efficiently. Subscriptions are compressed into a tree of object key pathways, traversed at `O(log n)`. Our canvas example is able to handle massive grids of pixels:

screen shot 2017-10-10 at 8 16 42 am

**So why is it getting moved out of core**

Usage in Microcosm was never publicly documented, and no internal pieces rely on it. It is dead weight. 

As we explore a querying interface for Microcosm, the CompareTree might get more attention, however I want to separate it from Microcosm core in the interim to free up some design decision making.

Long term, I still want to use something _like_ CompareTree. I'm particularly interested in how it could be blended with GraphQL queries. 

**This is a breaking change, and marks the first step towards 13.x**


packages/microcosm-examples/canvas/app/boot.js
```diff
@@ -1,13 +1,17 @@
...
+let tree = new CompareTree()
+
+repo.on('change', tree.update, tree)
 
 let size = 20
 let limit = Infinity
 let rows = Math.min((window.innerWidth / size) | 0, limit)
 let columns = Math.min((window.innerHeight / size) | 0, limit)
-let writes = size * 25
+let writes = size * 50
```
nhunzaker: Something about this change or a prior change caused a big perf improvement. I wish I knew what it was. I'll make a note to dig into this.

mackermedia:
👍  LGTM 🌮

Issue #427, opened 2017-09-26
Breaking change: Microcosm forks can not add domains at the same key, only one domain per key


nhunzaker:
Generally I want to make Domain mounting more strict so that I can add some optimizations and support new features. For some new features I need isolation between domains. Sharing keys, mounting a domain to the same key as a parent, and root level Domains are problematic for those use cases.

## No duplicate domain keys on forks

While I like forks for mounting new state, I think I was wrong to allow them to process data again:

```javascript
let parent = new Microcosm()
parent.addDomain('planets', {...})

let child = parent.fork
child.addDomain('planets', {...}) // Add new processors on top of whatever the parent gives you
```

The original idea was to use this for features like pagination and layering on app section specific data to a resource. It turns out that this just makes things complicated and our presenter models are more appropriate for that.

It also causes a lot of internal complexity and extra object clones, and I'd appreciate not having to maintain that.

## No duplicate domain keys on the same repo

Additionally, I'd like to make it so that you can only mount one domain per key. Adding more than one domain for the same key has only caused trouble. One alternative would be to allow you to add "subdomains" at specific keys within a root level domain.

## No root level domains

I want to get rid of root level domains as a part of this purge as well. They prevent us from doing some neat optimizations and other future ideas for domains:

```
this.addDomain([], RootDomain)
```

PR #435, opened 2017-10-12
Repos and Forks can not add multiple domains for the same key

nhunzaker:
This commit removes the ability to add multiple domains for the same key. This allows us to:

1. Remove the meta-domain
2. Remove the ADD_DOMAIN lifecycle action
3. Avoid merging data to avoid "spilling" keys

**Breaking changes**

1. Microcosms can no longer implement `register()`. This was intended to make it easier to build a Microcosm, but it is never used in real-world apps™️ and is problematic for future work.
2. Forks can not add domains on top of parents
3. Microcosms can not add multiple domains for the same key

Fixes https://github.com/vigetlabs/microcosm/issues/427.


packages/microcosm/src/domain-engine.js
```diff
@@ -76,28 +69,50 @@ class DomainEngine {
...
+    this.lifecycle[PATCH.toString()].push({
+      key: key,
+      scope: domain,
+      local: true,
+      steps: [
+        (state: *, data: any) => {
+          return data[key] !== undefined ? data[key] : state
+        }
+      ]
+    })
```
efatsi: can you explain what's up with `RESET.toString()` and `PATCH.toString()`? And what these two pushes do?

packages/microcosm/src/domain-engine.js
```diff
@@ -112,60 +127,78 @@ class DomainEngine {
...
-      } else {
-        // No: use the existing snapshot value (memoizing the domain handler)
-        result = set(result, key, get(snapshot.next, key))
+        // Recalculate state from the last answer
+        next = reduce(steps, action.payload, last, scope)
+      }
+
+      if (answer === state && next !== answer[key]) {
+        answer = clone(state)
       }
```
efatsi: can you summarize what this conditional is checking, and then doing if it returns true?

nhunzaker: Yep! This check is saying "If the answer object reference is the same as the state object reference, and the value has changed, then clone the answer".

This is pretty low level. Basically I want to avoid shallow cloning the next state object every time a domain handler runs.

packages/microcosm/src/domain-engine.js
nhunzaker: Yep. 

`PATCH.toString()` is to tell Flow that the underlying value is a string. This is _lame_.

These two pushes replace the MetaDomain as the method of modifying data after `repo.reset` or `repo.patch` fire.

For reset:

Update the value for the key the domain is managing if the key is defined in the data payload. Otherwise fall back to the initial state. So `repo.reset()` reverts to initial state, but `repo({ key: 'value' })` sets a specific value.

For patch:

The same is true, however the fallback is the _current_ value, rather than the initial state.

PR #436, opened 2017-10-13
Remove ability to reset Microcosm on construction

nhunzaker:
This commit removes support for the ability to reset state when a Microcosm is created. 

```javascript
let repo = new Repo({}, {
  planets: [ 'earth', 'mars', 'jupiter']
})
```
You mostly see this in tests, and it looks goofy. It's also easy to miss the third argument, which tells Microcosm to deserialize the data (like take string JSON and turn it into an object). Beyond that, there are a couple of other reasons:

1. I only want one way to reset state
2. It can be confusing to look at constructors that include seed data, particularly when deserialized
3. This is part of an effort to reduce some extraneous use cases to improve learnability/useabillity
4. You can't listen to failure (`repo.reset().onError(...)`)

I also took at pass at improving some tests around instantiation options while I was at it.

### DIFF
--- a/packages/microcosm/CHANGELOG.md
+++ b/packages/microcosm/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Moved internal state tree diffing tool to addon. 
 - Removed support for root level domains, forks can not add keys managed by parents. Repos can not add domains to the same key name.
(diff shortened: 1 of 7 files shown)


efatsi:
👍

PR #437, opened 2017-10-16
History changeset

nhunzaker:
**What**

Microcosm determines application state by taking a list of actions (history) and using their results to calculate a new state. When an action changed, History would emit a change event for that action and every action that followed it in this list.

This is a lot of events! Instead, this commit changes the strategy so that it only emits the _range_ of actions; the start and the end.

**Why**

Two reasons:

1. I want to eventually replace event emitters with [observables](https://github.com/tc39/proposal-observable), which more accurately match how actions work. This gets rid of one of those events "update" event, meaning we only need to figure out how to handle "remove", then we can drop the event emitter abstraction.
2. Performance. Because a Microcosm is responsible for iterating through actions instead of history, it can stop processing data when an action results in the same state. For long history chains, and for forks, this can avoid a lot of extra work.

### DIFF
--- a/packages/microcosm/src/microcosm.js
+++ b/packages/microcosm/src/microcosm.js
@@ -417,6 +417,22 @@ class Microcosm extends Emitter {
       status: action.status,
       payload: action.payload
     }
+
+    return snap.next !== next
+  }
+
+  _reconcileChangeset(action: Action, changeset: Action[]) {
+    for (var i = 0, len = changeset.length; i < len; i++) {
+      var changed = this._updateSnapshot(changeset[i])
+
+      if (changed === false) {
+        break
+      }
+    }
+
+    if (this.effects) {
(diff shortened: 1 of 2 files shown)


packages/microcosm/src/microcosm.js
```diff
@@ -417,6 +417,25 @@ class Microcosm extends Emitter {
...
+      if (changed === false || focus === end) {
+        break
+      }
+
+      focus = focus.next
+    }
+
+    if (this.effects) {
+      this._dispatchEffect(source)
+    }
```
efatsi: awesome

nhunzaker:
Yeah. Dropping support for sharing keys between parents and forks enables so many efficiencies and reductions in edge cases. I think we can even get rid of the `start/birth` actions. Maybe a followup PR.

PR #440, opened 2017-10-18
Batch remove action snapshots

nhunzaker:
**What**

This commit changes the way action snapshots are removed such that they are deleted in bulk. Additionally it moves the archive process to after the release process ends, preventing archiving from running multiple times between a release.

It also removes the BIRTH action, which was included to ensure that an action always had a parent. We don't need this. The only time an action will not have a parent is during START.

**Why**

Mostly just a performance optimization that was uncovered with #438 

---

Fixes #440

### DIFF
--- a/packages/microcosm/src/action.js
+++ b/packages/microcosm/src/action.js
@@ -38,13 +38,14 @@ class Action extends Emitter {
   children: Action[]
   revisions: Revision[]
   origin: Microcosm
+  connected: boolean
 
   constructor(command: string | Command, status: ?Status, origin: Microcosm) {
     super()
 
-    this.id = uid('action')
     this.command = tag(command)
     this.status = 'inactive'
+    this.id = uid(this.toString())
     this.payload = undefined
     this.disabled = false
     this.complete = false
@@ -54,6 +55,7 @@ class Action extends Emitter {
     this.children = []
     this.revisions = []
     this.origin = origin
+    this.connected = true
 
     if (status) {
       this._setState(status)
@@ -216,10 +218,6 @@ class Action extends Emitter {
     }).then(pass, fail)
   }
 
-  isDisconnected(): boolean {
-    return !this.parent
-  }
-
   /**
    * Remove the grandparent of this action, cutting off history.
    */
@@ -264,10 +262,18 @@ class Action extends Emitter {
    * from history.
    */
   remove() {
+    console.assert(this.connected, 'Action has already been removed.')
+
+    this.connected = false
+
+    this.children.length = 0
+
     if (this.parent) {
       this.parent.abandon(this)
-    } else {
-      console.assert(false, 'Action has already been removed.')
+    }
+
+    if (this.next) {
+      this.abandon(this.next)
     }
 
     this.removeAllListeners()
--- a/packages/microcosm/src/history.js
+++ b/packages/microcosm/src/history.js
@@ -10,7 +10,7 @@ import Action from './action'
 import Emitter from './emitter'
 import defaultUpdateStrategy from './default-update-strategy'
 import { merge } from './utils'
-import { BIRTH, START } from './lifecycle'
+import { START } from './lifecycle'
 import { type Updater } from './default-update-strategy'
 import { iteratorTag } from './symbols'
 
@@ -134,7 +134,7 @@ class History extends Emitter {
 
         // Ignore certain lifecycle actions that are only for
         // internal purposes
-        if (next && (next.command === BIRTH || next.command === START)) {
+        if (next && next.command === START) {
           return iterator.next()
         }
 
@@ -205,11 +205,6 @@ class History extends Emitter {
     if (this.head) {
       this.head.lead(action)
     } else {
-      // Always have a parent node, no matter what
-      let birth = new Action(BIRTH, 'resolve')
-
-      birth.adopt(action)
-
       this.root = action
     }
 
@@ -230,7 +225,7 @@ class History extends Emitter {
    * together to bridge the gap.
    */
   remove(action: Action) {
-    if (action.isDisconnected()) {
+    if (action.connected === false) {
       return
     }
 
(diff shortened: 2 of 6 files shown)


packages/microcosm/src/action.js
```diff
@@ -54,6 +55,7 @@ class Action extends Emitter {
     this.children = []
     this.revisions = []
     this.origin = origin
+    this.connected = true
```
nhunzaker: I didn't like how we determined if an action was disconnected or not based on the presence of `next` or `parent`.

packages/microcosm/src/history.js
```diff
@@ -336,13 +332,15 @@ class History extends Emitter {
...
 
-      this._emit('remove', root.parent)
-
       if (root.next) {
         root = root.next
       }
     }
 
+    if (root.parent) {
+      this._emit('remove', root.parent.parent, root.parent)
```
nhunzaker: This feels a little goofy. When we archive, we need to remove everything before the parent. Maybe this should just be `this._emit('remove', null, root.parent)`

packages/microcosm/src/action.js
```diff
@@ -264,10 +262,18 @@ class Action extends Emitter {
    * from history.
    */
   remove() {
+    console.assert(this.connected, 'Action has already been removed.')
+
+    this.connected = false
+
+    this.children.length = 0
```
efatsi: javascript please

efatsi: is `this.children = []` just as performant? because i think it reads wayyyy better

packages/microcosm/src/history.js
```diff
@@ -134,7 +134,7 @@ class History extends Emitter {
 
         // Ignore certain lifecycle actions that are only for
         // internal purposes
-        if (next && (next.command === BIRTH || next.command === START)) {
+        if (next && next.command === START) {
```
efatsi: how would the `next` of any action be `START`?

packages/microcosm/src/history.js
```diff
@@ -336,13 +332,15 @@ class History extends Emitter {
...
 
-      this._emit('remove', root.parent)
-
       if (root.next) {
         root = root.next
       }
     }
 
+    if (root.parent) {
+      this._emit('remove', root.parent.parent, null)
```
efatsi: the combination of `this._emit('remove', action, action)` and `this._emit('remove', root.parent.parent, null)` seems like a pretty funky code smell

packages/microcosm/src/microcosm.js
```diff
@@ -423,21 +423,30 @@ class Microcosm extends Emitter {
...
+    let focus = start
+    while (focus) {
+      delete this.snapshots[focus.id]
+
+      if (stop != null && focus !== stop) {
+        break
+      }
+
+      focus = focus.parent
+    }
```
efatsi: this is quite confusing. so it starts at an action, and moves up until it's passed the `stop`? unless there isn't a stop in which case it moves up until it hits the end of the chain?

nhunzaker: Yeah.

packages/microcosm/src/history.js
nhunzaker: Hah. Yeah.. you're right...

efatsi: likely tied to the smelly implementation of removeSnapshotRange.

packages/microcosm/src/action.js
nhunzaker: Good question. It'll definitely allocate more memory. Basically I just want to purge children before an action goes away. Maybe we don't need to do this.

packages/microcosm/src/history.js
nhunzaker: Very good question.

packages/microcosm/src/action.js
```diff
@@ -38,13 +38,14 @@ class Action extends Emitter {
   children: Action[]
   revisions: Revision[]
   origin: Microcosm
+  connected: boolean
```
efatsi: i wonder if `connected` is the right variable name here. root of my suspicion here is that this means you can have a `connected` action node without any parent or children (which seems like what "connected" would imply). is something like "active" more appropriate?

packages/microcosm/src/microcosm.js
efatsi: so two things that feel weird here. 1 is that you're starting at a node and then moving up, where as `_updateSnapshotRange` (which seems like it should be very close in implementation) starts at a node and moves down. minor thing, but still a context switch to be made in reading both. 2. this `(stop != null && focus !== stop)`. negative switch statements are already tough, 2 of them compound.

Does History know what the range it wants to remove whenever it's emitting `remove`? If so, maybe we can change the arguments here to make things more clear here.

efatsi: something like this work (needs cleaning):
```js
_removeSnapshotRange(start: Action, stop: ?Action) {
  let focus = start
  
  if (stop == null) {
    delete this.snapshots[focus.id]
  } else {
    while (focus) {
      delete this.snapshots[focus.id]

      if (focus == stop) {
        break
      }
      
      focus = focus.next
    }
  }
}
```

packages/microcosm/test/unit/history/remove.test.js
```diff
@@ -98,16 +98,15 @@ describe('History::remove', function() {
...
       repo.append(function two() {}, 'resolve')
       repo.append(function three() {}, 'resolve')
 
       let root = history.root
-      let next = root.next
 
       history.remove(root)
 
-      expect(history.root.id).toBe(next.id)
+      expect(history.root.id).toBe(one.id)
```
efatsi: so the START action is being removed in this instance?

nhunzaker:
@efatsi I'm going to close this one out. I have some changes to history that invalidate this in some upcoming work. Will post soon :)

PR #444, opened 2017-11-01
set up page structure markup

chloehwang:
In this PR, I laid out skeleton markup for the page structure and did some light styling (mainly on `nav`, `main`, `footer`). I already set up a basic CSS system in an earlier commit. Below is a screenshot of how it looks so far:

screen shot 2017-11-01 at 1 48 49 pm


packages/microcosm-www-next/src/layouts/index.js
```diff
@@ -2,28 +2,37 @@ import React from 'react'
...
-          style={{
-            color: 'black',
-            textDecoration: 'none',
-          }}
-        >
-          Microcosm
-        </Link>
-      </h1>
+      <Link to="/">
+        <img src="./static/images/logo.png" alt="Microcosm" />
```
chloehwang: should I include `.screenreader-only` text for Microcosm, or is the `alt` enough?

packages/microcosm-www-next/src/pages/index.js
```diff
@@ -2,12 +2,32 @@ import React from 'react'
...
-  </div>
+  <section>
+    <div className="wrapper">
+      <h1>
+        <span>01.</span>
+        Rendering a View
+      </h1>
+
+      <div className="figure-container">
+        <figure className="figure -left">
```
chloehwang: does it make sense to use `figure` here?

nhunzaker:
@chloehwang If you don't mind, could you run this at the root of the project:

```
yarn format
```

That'll make CI happy :)

packages/microcosm-www-next/src/layouts/index.js
nhunzaker: Alt should be enough. cc @ten1seven right?

packages/microcosm-www-next/src/pages/index.js
leobauza: I think so

packages/microcosm-www-next/src/layouts/index.js
leobauza: Alt should be enough, though I am not sure for SEO if not having a title to the page is ok?

package.json
```diff
@@ -44,6 +44,12 @@
...
     "packages/microcosm",
-    "packages/*"
+    "packages/microcosm-devtools",
+    "packages/microcosm-examples",
+    "packages/microcosm-graphql",
+    "packages/microcosm-http",
+    "packages/microcosm-preact",
+    "packages/microcosm-react-router",
+    "packages/microcosm-www"
   ]
```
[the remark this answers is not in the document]
chloehwang: interesting... ohhh monorepo. Wait, do we need to add `microcosm-www-next`? 

And thanks for the CI fix!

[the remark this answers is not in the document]
nhunzaker: Ya! Not yet. It looks like the Gatsby CI doesn't like something about it. I'm going to ask their team about it. In the mean time, let's just treat the new site as an island.

packages/microcosm-www-next/src/layouts/index.js
ten1seven: The `alt` attribute is enough for accessibility but @leobauza is correct, not having an `<h1>` is not good for overall accessibility or SEO because it hurts document structure.

chloehwang: okay noted, I will wrap this in an `<h1>`

PR #445, opened 2017-11-01
styled text, footer; started laying out section

chloehwang:
- Finished styling all text
- Finished styling footer
- Started laying out elements in the main section; implemented spacing.

Below is how it looks like now:

screen shot 2017-11-01 at 6 21 36 pm

### DIFF
--- a/packages/microcosm-www-next/src/layouts/index.js
+++ b/packages/microcosm-www-next/src/layouts/index.js
@@ -26,7 +26,7 @@ const Footer = () => (
         <VigetLogo fill="blue" />
       </Link>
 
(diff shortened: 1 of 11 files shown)


packages/microcosm-www-next/src/layouts/index.js
```diff
@@ -26,7 +26,11 @@ const Footer = () => (
         <VigetLogo fill="blue" />
       </Link>
 
-      <Link to="http://code.viget.com/microcosm/" target="_blank">
+      <Link
+        to="http://code.viget.com/microcosm/"
+        className="footer__cta"
+        target="_blank"
```
leobauza: 🔤

Issue #428, opened 2017-09-28
Upstream Task: Migrate repo.parallel fixes from 12.12.2 to 12.13+


solomonhawk:
This is a reminder for us to upstream fixes introduced in 12.12.2 to newer versions of Microcosm. Specifically: we fixed `repo.parallel` so that it returns the payloads of the child actions in the `onDone` handler attached to `repo.parallel`.

- [x] 12.13.x
- [ ] 12.14.x
- [ ] Master

PR #448, opened 2017-11-09
Upstream repo.parallel fix

nhunzaker:
We encountered an issue on an app using 12.12.x where we needed the
payloads from `repo.parallel` when it completed. This commit upstreams
that change.

Partially fixes https://github.com/vigetlabs/microcosm/issues/428

### DIFF
--- a/packages/microcosm/docs/api/microcosm.md
+++ b/packages/microcosm/docs/api/microcosm.md
@@ -333,7 +333,9 @@ fork.push(getPeople)
 
 Create a new "group" action bound to the resolution of a list of
 actions. If all actions resolve or cancel, the group action will
-resolve. If any action is rejected, the group action fails:
++resolve. If any action is rejected, the group action fails. If all
+actions resolve, their respective payloads will be available in the
+`onDone` callback:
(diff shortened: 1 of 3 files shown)


packages/microcosm/docs/api/microcosm.md
```diff
@@ -333,16 +333,18 @@ fork.push(getPeople)
...
 let group = repo.parallel([
   repo.push(actionOne),
   repo.push(actionTwo)
 ])
 
-group.onDone(function () {
-  console.log('hurrah!')
+group.onDone(function ([answer1, answer2]) {
+  console.log('hurrah!', answer1, answer2)
 })
```
nhunzaker: Long term, I think it would be great to also support objects, like:

```
let group = repo.parallel({
  one: repo.push(actionOne),
  two: repo.push(actionTwo)
})

let { one, two } = await group
```

I'll file an issue for it.

nhunzaker: https://github.com/vigetlabs/microcosm/issues/449

PR #458, opened 2017-12-22
Ch/qa scrolling

chloehwang:
This PR does some initial QA on layout and scroll behavior. It:

- implements polyfill for IntersectionObserver
- makes some minor tweaks for Safari
- implements feature queries to substitute for `position: sticky` on IE


packages/microcosm-www-next/src/layouts/index.js
```diff
@@ -4,6 +4,7 @@ import Link from 'gatsby-link'
 import Helmet from 'react-helmet'
 import VigetLogo from '../components/viget-logo'
 
+import 'intersection-observer'
```
chloehwang: polyfill

packages/microcosm-www-next/src/pages/index.js
```diff
@@ -35,7 +49,9 @@ export default class IndexPage extends React.Component {
     )
 
     //start observing each graphic
-    this.graphics.forEach(graphic => observer.observe(graphic))
+    for (let i = 0; i < this.graphics.length; i++) {
```
chloehwang: changed this to for loop because it wasn't working in IE

PR #459, opened 2018-01-03
Ch/design tweak test

chloehwang:

- Bugfix
- Feature
- Code style update
- Refactor
- Build tools
- Documentation
- Other, please describe:

- Yes
- No

- [ ] All tests are passing
- [ ] `yarn pretty` has been run
- [ ] Commits reference open issues

**Additional Information**


packages/microcosm-www-next/src/components/side-nav.js
```diff
@@ -0,0 +1,59 @@
...
+                  {data[section].heading}
+                </a>
+              </li>
+            ))}
+          </ol>
+        </nav>
+      </aside>
+    )
+  }
+}
```
nhunzaker: What if you tracked the current section as component state? It could default to `0`. Instead of manually assigning the class, you could make line 45 read like:

```javascript
className={section === this.state.current ? '-active' : ''}
```

Then you avoid needing to select the elements on initial render too.

packages/microcosm-www-next/src/pages/index.js
```diff
@@ -57,9 +58,10 @@ export default class IndexPage extends React.Component {
   onIntersection = observed => {
     let entry = observed[0]
     let section = parseInt(entry.target.dataset.section)
+    let isIntersecting = entry.intersectionRatio >= 1.0
```
nhunzaker: Is the `1.0` here the same threshold as `observeOptions.threshold`?

packages/microcosm-www-next/src/components/side-nav.js
chloehwang: good call, I was able to use a state prop that already existed 👍🏼

packages/microcosm-www-next/src/pages/index.js
chloehwang: yes, I will save that in a var

PR #465, opened 2018-01-08
add icons, minify images

chloehwang:
- refactored `ToggleContainer` into it's own React component
- add the computer and planet icons that go in the button
- made the button icon colors change on scroll
- minified images
  
untitled


packages/microcosm-www-next/src/stylesheets/structure/_section.scss
```diff
@@ -151,30 +151,6 @@ $section-top-padding-desktop: $nav-height + 110px;
   .toggle-container.-desktop & {
     margin-bottom: rem(30);
   }
-
-  &.-top {
-    &::after {
-      content: ' Microcosm';
```
chloehwang: now this text switching is all happening inside `pages/index.js`

leobauza:
Question about images in general. Since this is so image heavy how are you handling preloading or lazy loading or anything like that?

chloehwang:
@leobauza yes very good thought, I will think about and find a solution for that

PR #468, opened 2018-01-09
added lazy loader for images

chloehwang:
Added this [react-lazysizes](https://www.npmjs.com/package/react-lazysizes) plugin and set some load styles.


packages/microcosm-www-next/src/images/browser-icon.js
```diff
@@ -7,7 +7,6 @@ const BrowserIcon = () => (
     height="24"
     viewBox="0 0 24 24"
     style={{ marginRight: '10px' }}
-    preserveAspectRatio
```
chloehwang: got rid of this because it was causing console errors and also not needed

PR #469, opened 2018-01-09
lazy load images

chloehwang:
Added this [lazysizes](https://github.com/aFarkas/lazysizes/) plugin and set some load styles.


packages/microcosm-www-next/src/images/browser-icon.js
```diff
@@ -7,7 +7,6 @@ const BrowserIcon = () => (
     height="24"
     viewBox="0 0 24 24"
     style={{ marginRight: '10px' }}
-    preserveAspectRatio
```
chloehwang: got rid of this because it was causing console errors and also not needed

chloehwang:
@nhunzaker `yarn deploy` seems to work, but now it's failing because it can't find the `react-lazysizes` library I installed. I do see it in my `node_modules` so not sure why it can't find?

nhunzaker:
Bummer. Looks like it might be a caching issue. I think the build caches based on the root package.json. This site is separate, so it might have gotten stuck.

I clicked "rebuild without cache". Let's see what happens. I'll also make a note to look at why this happened.

nhunzaker:
Ahhh. I see now. `react-lazysizes` depends on `lazysizes`, which isn't getting pulled in for some reason. Hmmm.

chloehwang:
@nhunzaker couldn't figure out why yarn was not installing that `lazysizes` dependency 😣 it wasn't a well maintained package anyways so I used something else

packages/microcosm-www-next/src/stylesheets/components/_lazy-load.scss
```diff
@@ -0,0 +1,9 @@
+.lazyload,
+.lazyloading {
+	opacity: 0;
+}
+
+.lazyloaded {
+	opacity: 1;
+	transition: opacity 0.3s;
```
leobauza: are these over indented?

chloehwang: doesn't look like it in my code, not sure why it's doing that....

PR #473, opened 2018-01-30
Ch/change breakpoint

chloehwang:
Changed the point at which the layout breaks into desktop view - instead of going into desktop at >=1280px, now breaks at >=960px. 

untitled 2


chloehwang:
@leobauza oh yeah good point, I will test after I merge

PR #478, opened 2018-02-08
microcosm-dom package

nhunzaker:
This commit makes a single package for microcosm's preact and react bindings: microcosm-dom. This allows us to share nearly all of the code between these two rendering frameworks. 

I'd like to share more of the tests, but that turned out to be pretty hard. If we did that, I'd probably use the Preact tests, which do not rely on enzyme (which I think is React only).

```
import { Presenter } from 'microcosm-dom/react'

React.render(<Presenter />, document.body)

// ...................

import { ActionButton } from 'microcosm-dom/preact'

Preact.render(<ActionButton>Click Me</ActionButton>, document.body)
```


packages/microcosm-dom/src/presenter.js
```diff
@@ -0,0 +1,172 @@
...
+      this.teardown(this.repo, this.props, this.state)
+
+      if (this.didFork) {
+        this.repo.complete()
+      }
+    }
+
+    render() {
+      if (Array.isArray(this.props.children)) {
+        return this.props.children[0] || null
```
nhunzaker: This is the only really nasty wart. Preact sets children to an array _always_, where as React does not. Both must return null from render (I think this is true of React, but definitely Preact).

React can also return an array from render as of 16.

PR #486, opened 2018-03-20
Add value prop to ActionForm

### LINKED ISSUE #429: microcosm-addons: manually pass payload to ActionForm
For controlled forms, it would be great if you could pass in the payload for the form to send:

```javascript
class ControlledForm extends React.Component {
  // ... controlled input stuff
  render() {
    return (
      <ActionForm action="thing" payload={this.state.formData}>
        <input type="text" onChange={this.onChange} />
      </ActionForm>
    )
  }
}
```

In this mode:

1. The `serializer` prop would **not** be called.
2. The `prepare` prop _would_ be called

nhunzaker:
This commit adds a value prop to the ActionForm component of microcosm-dom. When set, the `value` prop is always passed as the serialized value of the form. This is intended to better support controlled forms.

This also achieves a nice parity with `ActionButton`, which lets you set a similar `value` property as the callback value when the button is clicked.

```javascript
import { login } from '../actions/session'

class Login extends React.Component {
  state = {
    email: '',
    password: ''
  }

  render() {
    let { email, password } = this.state

    return (
      <ActionForm action={login} value={this.state}>
        <input type="email" name="email" value={email} onChange={this.setField} />
        <input type="password" name="password" value={password} onChange={this.setField} />
      </ActionForm>
    )
  }

  setField = event => this.setState({ [event.target.name]: [event.target.value] })
}
```

Fixes #429


nhunzaker:
Ack yes. Good call!

dce:
When you say

> When set, the value prop is always passed as the serialized value of the form.

Do you mean "the value prop is always passed *intead of* the serialized value of the form"?

nhunzaker:
@dce yes. That is what I meant to say.

nhunzaker:
@cwmanning Documentation added!

PR #487, opened 2018-03-20
Domains can be defined as observables

nhunzaker:
This commit allows a domain to be a simple observable. Observable domains do not have a ledger and can not respond to actions. Their primary use case is to synchronize state with some other source. Like:

1. Window history
2. Websocket state
3. Geolocation

For these cases, you mostly just want to push an action to synchronize state, but that's a ton of boilerplate. Instead, you could do:

```javascript
import { history } from 'push-state-lib'

repo.addDomain('location', new Observable(observer => {
  observer.next(history.getLocation())

  let listener = history.listen(location => observer.next(location)

  // The return value of an observable's callback is a clean-up function
  // sort of like Domain::teardown
  return () => listener.remove()
})

repo.domains.location.subscribe(location => {
  console.log("The next location is", location)
})
```


packages/microcosm/src/domain.js
```diff
@@ -126,10 +126,6 @@ export class Domain<State: mixed = null> extends Subject {
     return this.serialize(this.valueOf())
   }
 
-  valueOf() {
-    return this.payload
-  }
```
nhunzaker: I can remove this because it is implemented by `Subject`

packages/microcosm/src/subject.js
```diff
@@ -160,12 +164,13 @@ export class Subject {
   }
 
   static hash(obj: *): Subject {
+    let subject = new Subject()
+
     if (getObservable(obj)) {
-      return obj
+      obj.subscribe(subject)
+      return subject
     }
```
nhunzaker: This change is important because the return value was inconsistent. I'm a little bummed Flow didn't catch this. We can't return an `Observable` because they play by different rules than `Subject`

packages/microcosm/test/unit/history/archive.test.js
```diff
@@ -48,10 +48,10 @@ describe('History::archive', function() {
 
     repo.history.archive()
 
-    expect(`${repo.history.root}`).toBe('second')
+    expect(repo.history.root.toString()).toBe('second')
 
     // This is two because we checked out two earlier
-    expect(`${repo.history.head}`).toBe('second')
+    expect(repo.history.head.toString()).toBe('second')
   })
```
nhunzaker: String interpolated values call `valueOf` before `toString`. TIL.

packages/microcosm/src/microcosm.js
```diff
@@ -59,21 +60,25 @@ export class Microcosm extends Subject {
     // NOOP
   }
 
-  addDomain(key: string, blueprint: *, config?: *): Domain<*> {
```
dce: Is this (`<*>`) Flow?

nhunzaker: Ya. It's a generic. The type signature for a domain is Domain<State>. State can be defined as any value, it just has to be consistent across `getInitialState` and action callbacks. That way you can guarantee state shape stays the same. You can also get the type signature of a piece of state in Presenters :)

nhunzaker: Sort of long term dream is that this type signature can be extended to more powerful view modelling abstractions so you can get strong type safety writing view models for your UI stuff.

PR #499, opened 2018-04-10
WIP - Microcosm Agents

nhunzaker:
I sort of hesitate to send this out, but this work creates a common abstraction between Domains and Entities: the Agent.

An agent is the baseline implementation of an object that manages some state (as a class instance) and can receive messages from a Microcosm history.


nhunzaker:
No reason for this to depend on the json schema work. I'm going to pull this out into a separate PR focused just on agents.

PR #500, opened 2018-04-11
[microcosm] Unify effect and domain behavior through agent class

nhunzaker:
This commit adds a new class to describe the dispatch behavior: an agent. Agents receive action dispatches from history and can manage their own internal state:

```javascript
import { start, stop } from '../actions/gps'

class GPS extends Agent () {
  setup() {
    this.listen()
  }

  listen() {
    this.watcher = navigator.geolocation.watchPosition(this.next)
  }

  ignore() {
    navigator.geolocation.clearWatch(this.watcher);
  }

  receive(action) {
    switch(action) {
      case start: 
        return this.listen()
      case stop: 
        return this.ignore()
      default: 
        return null
    }
  }
}

let gps = new GPS()

gps.subscribe(next => console.log(next)) // location.... location .... location....

gps.send(start)
```

I still haven't figured out the public API for Agents yet, but I think they will be a useful escape hatch for creating new types of Domain/Effect-like abstractions. I'm also curious if they could be used as a type of viewmodel class, but I'm not quite there yet.

### DIFF
--- a/packages/microcosm/src/agent.js
+++ b/packages/microcosm/src/agent.js
@@ -1,0 +1,52 @@
+/**
+ * @fileoverview An agent is a general abstraction for intercepting actions.
+ * It is the basis of Domain and Effect
+ * @flow
+ */
+
+import { type Microcosm } from '../src/microcosm'
+import { Subject } from './subject'
+import { merge } from './data'
+
+export class Agent extends Subject {
+  repo: Microcosm
+  options: Object
+
+  static defaults: ?Object
+
+  constructor(repo: Microcosm, options?: Object) {
+    super(null, options)
+
+    this.repo = repo
+    this.options = merge(repo.options, this.constructor.defaults, options)
+
+    let tracker = repo.history.subscribe(this.receive.bind(this))
+
+    repo.subscribe({
+      complete: () => {
+        tracker.unsubscribe()
+        this.teardown(this.repo, this.options)
+      }
+    })
+
+    this.setup(this.repo, this.options)
+  }
+
+  /**
+   * Setup runs right after an agent is added to a Microcosm. It
+   * receives that repo and any options passed as the second argument.
+   */
+  setup(repo?: Microcosm, options?: Object): void {}
+
+  /**
+   * Runs whenever a Microcosm is torn down. This usually happens when
+   * a Presenter component unmounts. Useful for cleaning up work done
+   * in `setup()`.
+   */
+  teardown(repo?: Microcosm, options?: Object): void {}
+
+  /**
+   * Called whenever an agent receives a new action
+   */
+  receive(action: Subject): void {}
+}
--- a/packages/microcosm/src/domain.js
+++ b/packages/microcosm/src/domain.js
@@ -7,70 +7,30 @@ import { Registry } from './registry'
 import { EMPTY_OBJECT } from './empty'
 import { RESET, PATCH } from './lifecycle'
 import { Ledger } from './ledger'
+import { Agent } from './agent'
 
 type DomainHandler<State> = (state: State, payload?: *, meta?: *) => State
 
-export class Domain<State: mixed = null> extends Subject {
-  repo: Microcosm
-  options: Object
+type DomainRegistry<State> = {
+  [string]: DomainHandler<State>
+}
+
+export class Domain<State: mixed = null> extends Agent {
   _registry: Registry
   _ledger: Ledger<State>
 
-  static defaults: ?Object
-
-  static from(config: *): Class<Domain<*>> {
-    if (typeof config === 'function') {
-      return config
-    }
-
-    function NewEntity(repo: Microcosm, options?: Object) {
-      // $FlowFixMe
-      Domain.prototype.constructor.call(this, repo, options)
-    }
-
-    NewEntity.prototype = Object.create(Domain.prototype)
-
-    for (var key in config) {
-      NewEntity.prototype[key] = config[key]
-    }
-
-    // $FlowFixMe
-    return NewEntity
-  }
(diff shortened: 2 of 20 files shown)


packages/microcosm/src/subject.js
```diff
@@ -6,14 +6,14 @@ import { noop, EMPTY_SUBSCRIPTION } from './empty'
 import { set, merge } from './data'
 
 export class Subject {
-  meta: { tag: *, status: string, origin: Microcosm }
+  meta: { key: *, status: string, origin: Microcosm }
```
nhunzaker: I renamed `tag` to `key` for consistencies sake.

packages/microcosm/src/subject.js
```diff
@@ -134,14 +134,14 @@ export class Subject {
   }
 
   toString(): string {
-    return this.meta.tag || 'Subject'
+    return this.meta.key
```
nhunzaker: Feels good to get rid of this null state :)

packages/microcosm/test/helpers.js
```diff
@@ -4,7 +4,7 @@ export function asTree(history) {
   function dig(focus) {
     let children = focus.children.map(dig)
 
-    return children.length ? [focus.tag, ...children] : focus.tag
+    return children.length ? [focus.key, ...children] : focus.key
   }
 
   return asciitree(dig(history.toJSON().tree))
```
nhunzaker: As a side note, this is a utility for serializing history into a tree visualization. I'm using this for snapshot testing the history structure:

```
    a
    |
    b
   / \
  c   d
```

Super useful!

packages/microcosm/src/agent.js
```diff
@@ -0,0 +1,52 @@
...
+import { Subject } from './subject'
+import { merge } from './data'
+
+export class Agent extends Subject {
+  repo: Microcosm
+  options: Object
+
+  static defaults: ?Object
+
+  constructor(repo: Microcosm, options?: Object) {
```
nhunzaker: Thinking on this a bit more, I'm also curious about the idea of an `AgentPool` class that could distribute state management across a group of presenters by sharing agents with the same configuration.

packages/microcosm/src/domain.js
```diff
@@ -116,10 +62,22 @@ export class Domain<State: mixed = null> extends Subject {
...
+
+    if (next !== this.payload) {
+      this.next(next)
+    }
+
+    // TODO: This could probably be a generic storage solution
+    // that cleaned up keys as actions completed
+    this._ledger.clean(action)
+  }
+
```
dce: ❤️ how much this cleans up Domain.

packages/microcosm/src/agent.js
nhunzaker: You'd do that by deriving a hashcode from the repo's identity and values passed via options

packages/microcosm/src/proto.js
```diff
@@ -0,0 +1,23 @@
...
+    Object.defineProperty(NewEntity.prototype, key, {
+      configurable: true,
+      enumerable: false,
+      writeable: true,
+      value: config[key]
+    })
+  }
+
+  return NewEntity
+}
```
dce: Does this have a test?

nhunzaker: It gets covered through domain/effect usage in other tests, but thanks for calling it out. I should test to cover a few cases:

- `null`
- There are certain rules for [property descriptors of ES6 classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptors), I should make this play nice

Really, I don't want to have something like this in Microcosm, but we've allowed users to pass domains/effects as POJOs forever. I'm thinking about deprecating it, but actually enforcing that domains/effects need to inherit from a library class might need to go out in 14.x.

packages/microcosm/src/agent.js
```diff
@@ -0,0 +1,52 @@
...
+    this.repo = repo
+    this.options = merge(repo.options, this.constructor.defaults, options)
+
+    let tracker = repo.history.subscribe(this.receive.bind(this))
+
+    repo.subscribe({
+      complete: () => {
+        tracker.unsubscribe()
+        this.teardown(this.repo, this.options)
+      }
```
efatsi: looks like this is existing functionality from domains/effects, but what would cause a `complete` action?

efatsi:
+1 to the refactoring here. Personally, I'm not too interested in a public API for Agents - I can't think of a use case for cutting off your own class that extends Agent when you have Domains/Effects at your fingertips.

nhunzaker:
@dce Looks like I was missing coverage in a few areas. In 1e37997a I added a few more tests and added an assertion to warn users when they add a null domain/effect. This achieves complete coverage of the proto module.

nhunzaker:
@efatsi Agreed. I don't want to advertise it for the time being, but I _am_ quite curious about how we could use it to create domains that do not rely on history.

packages/microcosm/src/agent.js
nhunzaker: Microcosm's are Observables (implemented as Subjects ). When you call `complete()` on a Microcosm, it has the same effect as calling `shutdown` in Microcosm 12. Instead of subscribing to the `teardown` event using an event emitter, repos/domains/presenters now listen to the `complete` lifecycle event.

TBH, I think there's an improvement that could be made here. This feels like bad design. I wonder if I should give Subject itself a setup and teardown method. You wouldn't have to implement it, but it could be inherited through Agent to Domains and Effects.

nhunzaker:
Thanks everyone!

PR #505, opened 2018-04-17
[microcosm] Add remove data helper. Setting undefined removes props

nhunzaker:
This commit updates the way `set` works such that it removes a property if it is undefined. It also adds an `remove` function, which is an alias for `set(object, prop, undefined)`.

```javascript
import { set, remove } from 'microcosm'

val records = { 1: { name: 'Billy' }, 2: { name: 'Sal' } }

// These are the same
set(records, 1, undefined) // { 2: { name: 'Sal' } }
remove(records, 1)         // { 2: { name: 'Sal' } }
```

I also took this as an opportunity to rework flow types and refactor some of the data operation code to avoid null checks.


packages/microcosm/src/data.js
```diff
@@ -117,49 +132,52 @@ export function get(object: ?Object, keyPath: *, fallback?: *) {
 
 /**
  * Merge any number of objects into a provided object.
+ * $FlowFixMe This function enumerates on variable arguments
  */
-export function merge(...args: Array<?Object>): Object {
-  let copy = null
-  let subject = null
+export function merge(): Object {
```
nhunzaker: The update here: instead of `...spread`'ing arguments into an array, we just enumerate over the arguments list. This prevents an array allocation and extra enumeration over `arguments`.

packages/microcosm/src/data.js
```diff
@@ -117,49 +132,52 @@ export function get(object: ?Object, keyPath: *, fallback?: *) {
...
         }
 
         copy[key] = next[key]
       }
     }
   }
 
-  return copy || {}
+  return copy
 }
```
nhunzaker: Other update: `merge` is non-destructive. It returns a new object if any of the keys are different. Using a boolean flag to determine if we've already copied, and tweaking the guard on line 148, allowed me to get rid of the null state.

efatsi: any benefit to test this update?

nhunzaker: I had this before because I couldn't convince flow that I wasn't returning null, since `copy` was initialized as `null`.

However, thinking it more, we shouldn't ever return EMPTY_OBJECT because it is sealed and a user might try to add a key to it assuming it's "save".

So I should cover that case, for sure.

packages/microcosm/src/data.js
```diff
@@ -117,49 +129,48 @@ export function get(object: ?Object, keyPath: *, fallback?: *) {
...
+          cloned = true
         }
 
         copy[key] = next[key]
       }
     }
   }
 
-  return copy || {}
+  return copy == EMPTY_OBJECT ? {} : copy
```
nhunzaker: @efatsi I ended up needing this check, but we now have coverage for it. 

Logically speaking, `EMPTY_OBJECT` is the same as `null` in this case. It's just that it isn't actually null. It's a null object I guess. I like this better, and it makes Flow happier.

What do you think?

efatsi: no strong opinions here from me

efatsi: no strong opinions here from me

nhunzaker: Cool. I'll ship this. Thanks!

PR #508, opened 2018-04-23
Fix edge case with patch/reset. Add dispatch optimization.

nhunzaker:
Adds two fixes/optimizations:

**[microcosm] Patch and Reset only trigger on complete**

This commit fixes a bug where patch and reset would trigger on every
state update. I only noticed this after doing some work on a scheduler
for Microcosm, which has some improvements in error handling.

**[microcosm] Do not process actions that are only started**

This commit commit adds an optimization to domain dispatch that
prevents the ledger from assigning state for actions that have only
started, but have no progress.


packages/microcosm/src/domain.js
```diff
@@ -85,10 +88,16 @@ export class Domain<State: mixed = null> extends Agent {
   }
 
   _resolve(action: Subject): DomainHandler<State>[] {
-    switch (action.toString()) {
-      case String(RESET):
-      case String(PATCH):
```
nhunzaker: Doing some benchmarks on my scheduler branch, wrapping a String like this was a major contributor to JS execution time.

packages/microcosm/src/domain.js
```diff
@@ -129,8 +137,22 @@ export class Domain<State: mixed = null> extends Agent {
...
-    let value = key in data ? data[key] : this.getInitialState()
+    console.assert(
+      meta.status === 'complete',
+      'Unable to reset or patch from incomplete action. This is an internal Microcosm error.'
+    )
+
+    console.assert(
+      data,
+      'Unable to reset or patch, no data provided. This is an internal Microcosm error.'
+    )
```
nhunzaker: These assertions were the best way to get test coverage on this case. Otherwise I have to dig into Microcosm internals in my tests, which is problematic for testing the production build.

nhunzaker:
Thanks, @zporter!

PR #509, opened 2018-04-24
Refine microcosm-http. and microcosm-dom Add file-uploader example

nhunzaker:
This commit moves the file-upload example from `microcosm-http` over to the examples directory and updates it with a cleaner interface.

In the process of doing this, I made a few improvements that I've pointed out within the file diff.

---

uploader


packages/examples/file-uploads/src/views/file-uploader.js
```diff
@@ -0,0 +1,46 @@
...
+
+    return (
+      <ActionForm
+        action={uploadFile}
+        serializer={asFormData}
+        queue={this.queue}
+        onSend={this.trackProgress}
+      >
+        <Progress status={status} file={file} onCancel={this.queue.clear} />
+
```
nhunzaker: New here: you can pass a `Subject` to `ActionForm` and `ActionButton`. This lets you control the underlying scheduling.

New here: `Subject:clear()`. This lets you cancel all subscriptions but keep the subject open. It's sort of like a reset.

packages/examples/file-uploads/src/views/file-uploader.js
```diff
@@ -0,0 +1,46 @@
...
+          <button>Upload</button>
+        </footer>
+      </ActionForm>
+    )
+  }
+
+  trackProgress = action => {
+    action.every(iteration =>
+      this.setState({ status: iteration.status, file: iteration.payload })
+    )
```
nhunzaker: New here: `action.every` is a subscription callback that fires on every status change.

packages/microcosm-dom/src/action-button.js
```diff
@@ -15,7 +15,7 @@ export function generateActionButton(createElement, Component) {
     }
 
     componentWillUnmount() {
-      this.queue.cancel()
+      this.queue.complete()
```
nhunzaker: I don't want to cancel. That would cancel any outstanding actions, which might not be what the user wants. If that is the desired behavior, you can now do that yourself.

packages/microcosm-dom/src/action-form.js
```diff
@@ -43,21 +43,19 @@ export function generateActionForm(createElement, Component) {
 
     submit(event) {
       let result = this.send(this.props.action, this._parameterize())
+      let action = Subject.hash(result)
```
nhunzaker: Wrapping the result in `Subject.hash` gives me either:

1. The same action back
2. A `Subject` version of whatever send returned. 

That means you can pass anything into the `send` prop and the action form will turn it into the same interface as an action (Subjects). You'll always get a Subject back from send now, every time.

So you don't really need a repo. Just a send method:

```javascript
let send = () => Promise.resolve(true)
let callback = (subject) => {}

<ActionButton send={send} onSend={callback}/>
```

packages/microcosm/test/unit/subject.test.js
```diff
@@ -168,6 +168,30 @@ describe('Subject', function() {
     })
   })
 
+  describe('clear', function() {
+    it('resets the usbject', function() {
```
zporter: typo: `subject`

packages/examples/file-uploads/src/views/file-uploader.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import { uploadFile } from '../actions/files'
+import { Subject } from 'microcosm'
+import { Presenter, ActionForm } from 'microcosm-dom'
+import { Progress } from './progress'
+
+function asFormData(form) {
+  return { data: new FormData(form) }
```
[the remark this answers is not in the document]
nhunzaker: I think a part of XMLHttpRequest 2.0:

> The FormData interface provides a way to easily construct a set of key/value pairs representing form fields and their values, which can then be easily sent using the XMLHttpRequest.send() method. It uses the same format a form would use if the encoding type were set to "multipart/form-data".

https://developer.mozilla.org/en-US/docs/Web/API/FormData

packages/microcosm-dom/test/presenter.test.js
```diff
@@ -540,7 +540,7 @@ describe('::teardown', function() {
     expect(renders).toBe(1)
   })
 
-  it('changes during teardown do not cause a recalculation', function() {
+  it('changes during teardown do not cause a recalculation', async function() {
```
zporter: total nitpick, but this might read better as:
```js
it('does not recalculate when changes during teardown occur'
```

packages/microcosm/test/unit/subject.test.js
efatsi: usbject

efatsi:
Cool stuff, like the way that `click` and `submit` have simplified down.

packages/examples/file-uploads/src/views/file-uploader.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export class FileUploader extends Presenter {
+  state = {
+    status: 'inactive',
+    file: undefined
+  }
+
+  queue = new Subject()
```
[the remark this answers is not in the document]
nhunzaker: I'm using it waaay more than Observable. It's a way to get the observable API externally, like this, but inside out:

```
new Observable(observer => {
  observer.next(1)
  observer.next(2)
  observer.next(3)
  observer.complete()
})
```

Can be:

```
let subject = new Subject()

subject.next(1)
subject.next(2)
subject.next(3)
subject.complete()
```

The advantage of an observable is that the behavior inside is lazy based on if anything subscribes to it. Subjects are more active.

packages/microcosm/test/unit/subject.test.js
nhunzaker: 😨

packages/microcosm-dom/test/presenter.test.js
nhunzaker: Great suggestion!

packages/examples/file-uploads/src/views/file-uploader.js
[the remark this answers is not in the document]
efatsi: Ahh, is it naive to think that this would be a reasonable default for the `serializer` ActionForm property?

packages/examples/file-uploads/src/views/file-uploader.js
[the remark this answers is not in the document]
efatsi: That's awesome. Not sure if `Subject` is already in your crosshairs for "needs a better name", but it's quite vague at the moment.

packages/examples/file-uploads/src/views/file-uploader.js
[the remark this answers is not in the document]
nhunzaker: Actually I was just typing this up. I'd be okay with dropping `form-serialize` and just making this the default. I wonder what turning FormData into JSON would look like.

packages/examples/file-uploads/src/views/file-uploader.js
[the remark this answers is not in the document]
nhunzaker: It is. It comes from Observables literature, but I agree that it's vague. I'd be happy to figure out another name.

Working more on scheduler stuff, I have some times wondered about calling this `Job` or `Process`. But I dunno.

packages/examples/file-uploads/src/views/file-uploader.js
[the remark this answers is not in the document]
nhunzaker: Issue filed:
https://github.com/vigetlabs/microcosm/issues/510

packages/examples/file-uploads/src/views/file-uploader.js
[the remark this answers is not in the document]
nhunzaker: Issue filed:
https://github.com/vigetlabs/microcosm/issues/511

packages/microcosm-dom/test/presenter.test.js
nhunzaker: Done in 0e03b9ce

packages/microcosm/test/unit/subject.test.js
nhunzaker: Done in 0e03b9ce

nhunzaker:
Thanks, guys!

PR #512, opened 2018-04-25
[microcosm] Async observables

nhunzaker:
*Review notes:* It's probably best to view this in split-view

### What is it

This commit updates the Observable abstraction in Microcosm such that it is asynchronous.

Updates to observables are queued with a universal scheduler. Until this scheduler "flushes" an observable, updates are buffered within an internal list of updates.

### Why 

The scheduler dispatches updates as fast as the browser will allow it to, using requestIdleCallback. When requestIdleCallback is not provided, it uses setTimeout.

This means that Microcosm's scheduler does not block the browser. Work might be slow, but the browser never suffers.


packages/microcosm/src/scheduler.js
```diff
@@ -0,0 +1,106 @@
...
+
+    this.scheduled = true
+    this.method(this.tick)
+  }
+
+  tick(idleDeadline) {
+    try {
+      while (this.queue.length && idleDeadline.timeRemaining() > 0) {
+        this.queue.shift()()
+      }
```
nhunzaker: Using [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback)'s API, this says:

> Run as many queued jobs as possible until the browser says it has to do something, like paint.

packages/microcosm/src/scheduler.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  flush() {
+    this.tick(untilEmpty)
+  }
+
+  onError(fn) {
+    this.errorCallbacks.push(fn)
+  }
```
nhunzaker: This lets you subscribe, globally, to uncaught errors executing jobs. Jobs is vague here. This is basically any state update within a Microcosm, Domain, Effect, or Action. Period.

Pretty neat :)

packages/microcosm/test/setup.js
```diff
@@ -1,6 +1,10 @@
-import { get } from 'microcosm'
+import { get, scheduler } from 'microcosm'
 import expect from 'expect'
 
+scheduler().onError(error => {
+  throw error instanceof Error ? error : new Error(error)
+})
```
nhunzaker: Says: while testing, if a job has an uncaught exception, fail immediately.

packages/microcosm/test/unit/observable.test.js
```diff
@@ -14,10 +14,28 @@ describe('Observable', function() {
...
+        throw 'Terrible things'
+      })
+
+      observable.subscribe({
+        error(reason) {
+          expect(reason).toEqual('Terrible things')
+        }
+      })
+
+      scheduler().flush()
```
nhunzaker: Flush is a nice utility for forcing all updates to send out. You could also use async/await like `await scheduler`, but this has the nice benefit of test execution time.

packages/microcosm/src/observable.js
```diff
@@ -181,105 +159,136 @@ export class Observable {
...
+function flushSubscription(subscription: Subscription): void {
+  let queue = subscription._queue
 
-  observer.complete()
+  subscription._queue = []
+  subscription._state = WORKING
 
-  cleanupSubscription(subscription)
-}
+  for (let i = 0; i < queue.length; ++i) {
```
zporter: would `++i` increment `i` to `1` before the block is called?

packages/microcosm/src/scheduler.js
```diff
@@ -0,0 +1,106 @@
...
+    )
+
+    this.queue.push(cb)
+    this.schedule()
+  }
+
+  schedule() {
+    if (this.scheduled) {
+      return
+    }
```
zporter: nice catch 👏

packages/microcosm/src/observable.js
nhunzaker: Nope! The third statement is the post-condition of the loop. This is just confusing. I'll replace it with `i++`, which I should have had to begin with.

packages/microcosm/test/unit/observable.test.js
```diff
@@ -14,10 +14,28 @@ describe('Observable', function() {
           expect(reason).toEqual('Terrible things')
         }
       })
+
+      scheduler().flush()
+    })
+
+    it('cleans ups any enqueued tasks on unsubscribe', () => {
```
zporter: nitpick: `cleans up`

nhunzaker: I keep all my ups clena.

nhunzaker:
Thanks, @zporter !

PR #513, opened 2018-04-25
[microcosm] Async, scheduled, subjects

nhunzaker:
This commit brings the asychronous behavior of Observables over to
Subjects. It is also able to consolidate some behavior by extending
Observable.

The async behavior of Subjects is different in that they track a
present value. When subscribed to, Subjects immediately emit an
update. All future work is scheduled.

This is a really nice system because it allows Presenter view models
to immediately have a value, and for state to be synchronized by the
time Effects dispatch.


packages/microcosm/src/subject.js
```diff
@@ -158,87 +137,151 @@ export class Subject {
...
+    if (jobs <= 0) {
+      subject.complete()
+    }
 
-    function complete() {
-      if (--jobs <= 0) {
-        subject.complete()
-      }
+    return subject
+  }
```
nhunzaker: This `.hash` static method is crazy big. It really needs to be its own class.

packages/microcosm/src/subject.js
```diff
@@ -1,140 +1,119 @@
...
+  type SubscriptionObserver,
+  type Cleanup
+} from './observable'
 import { observable } from './symbols'
-import { noop, EMPTY_SUBSCRIPTION } from './empty'
 import { set, merge } from './data'
+import { isPromise, isObservable, isObject, isPlainObject } from './type-checks'
 
-export class Subject {
+export class Subject extends Observable {
```
nhunzaker: A really nice side-effect of extending from Observable is that the scheduler and helper methods like `.map` and `.subscribe` can be shared. It truly is an extension of Observable now.

packages/microcosm/src/subject.js
```diff
@@ -158,87 +137,151 @@ export class Subject {
...
+    .then(subject.next)
+    .then(() => subject.complete())
+    .catch(subject.error)
+
+  return subject
+}
+
+function fromObservable(observable: Observable): Subject {
+  let subject = new Subject()
+  observable.subscribe(subject)
```
zporter: does this mutate the `subject` in place?

nhunzaker: It places the subject downstream to the observable. Because Subjects adhere to the subscriber interface, they can  be passed straight into subscribe. This is the same as:

```javascript
observable.subscribe({
  next: value => subject.next(value),
  error: reason => subject.error(reason),
  complete: () => subject.complete()
  cancel: () => subject.cancel()
})
```

nhunzaker: I should make an interface for this actually. Both Subscription and Subject should implement it. Jotting that down...

packages/microcosm/src/subject.js
nhunzaker: I did this in https://github.com/vigetlabs/microcosm/pull/514

Issue #511, opened 2018-04-24
[microcosm] Finalize terminology


nhunzaker:
Schedulers, observables, subjects? We need to make sure everything is clear and documentation has been updated before we send out 13.x.

Follow-up from https://github.com/vigetlabs/microcosm/pull/509#discussion_r183797074

PR #514, opened 2018-04-26
[microcosm] Split Subject.hash into separate class

nhunzaker:
But it was getting pretty massive, and I didn't like having it as a static method on Subject. This commit moves this function into a separate class that extends from Subject.

## What is Subject.hash

Subject.hash is a method used in a few places to reduce an object of nested asynchronous values into a POJO. This is extremely useful for building view models and for generally reducing down any complex async type in Microcosm down into something usable to the outside world:

```javascript
import { http } from 'microcosm-http'

let answer = Subject.hash({
  users: http.get('/users'),
  postsWithComments: Subject.hash({
    posts: http.get('/posts'),
    comments: http.get('/comments')
  }).map(mergePostsWithComments)
})

// This will emit a stream of updates as it gets answers
answer.subscribe({
  next() {}, // Partially loaded. You might have users, but not posts
  complete() { }, // All users, posts, and comments are loaded
  error () {} // Something went wrong with this entire behavior
})

// Or say you stop caring:
// with microcosm-http, this would cancel all requests
answer.cancel()
```

_This request behavior is additionally illustrated in the [file-uploads example](https://github.com/vigetlabs/microcosm/blob/master/packages/examples/file-uploads/src/views/file-uploader.js)._

## Why call it SubjectHash?

I don't know. I think this is the closest thing we have to the concept of a ViewModel, but it covers a broad use case of behavior. SubjectTree also came to mind, or something that doesn't use the word Subject, which is pretty vague. I have an issue to track that in https://github.com/vigetlabs/microcosm/issues/511


nhunzaker:
@zporter Renamed this to SubjectMap. I'm going to keep jamming on a name, but I'm hopeful that I'll hit some clarity once I've better distinguished the responsibility of this component in Presenters vs Microcosm itself.

Thanks for the review!

PR #516, opened 2018-05-08
[microcosm] Always, synchronously, emit the first observer.next()

nhunzaker:
This PR updates Observable such that the first "next" callback invokes immediately. This is important for Observable::map, particularly when binding data in React. Otherwise, React renders before the scheduler clears the queue for an observable. So you'll end up with something like:

```javascript
class UserShow extends Presenter {
  getModel(repo, { id }) {
    let { users } = repo.domains

    return {
      user: users.map(users => find(users, { id })
    }
  }
  render() {
    console.log(this.model.user) 
    // Always undefined initially because the observable callback is still
    // in the scheduler
  }
}
```

Essentially, this makes it so that you can always get a value immediately, even if the initial payload is empty.

### DIFF
--- a/packages/microcosm/src/observable.js
+++ b/packages/microcosm/src/observable.js
@@ -9,14 +9,16 @@ import { noop } from './empty'
 import { scheduler } from './scheduler'
 import { isObject } from './type-checks'
 
+// A subscription is constructing
+const START = 0
 // A subscription has no work in progress
-const IDLE = 0
+const IDLE = 1
 // A subscription will send, but the scheduler is not ready for it.
-const BUFFERING = 1
+const BUFFERING = 2
 // A subscription is processing
-const WORKING = 2
+const WORKING = 3
 // A subscription is done
-const CLOSED = 3
+const CLOSED = 4
 
 const NEXT = 'next'
 const ERROR = 'error'
@@ -82,7 +84,7 @@ export class Subscription implements Unsubscribable {
   _queue: Queue
 
   constructor(observer: Observer, subscriber: Subscriber, origin: *) {
-    this._state = IDLE
+    this._state = START
     this._cleanup = noop
     this._queue = []
     this._observer = observer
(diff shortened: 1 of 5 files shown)


packages/microcosm/src/observable.js
```diff
@@ -94,9 +96,13 @@ export class Subscription implements Unsubscribable {
       this._cleanup = subscriber.call(origin, subscriptionObserver) || noop
     } catch (error) {
       subscriptionObserver.error(error)
+
+      if (observer.error === noop) {
+        scheduler().raise(error)
+      }
```
nhunzaker: New: If an exception is raised in a subscriber and there isn't an error handler, we raise with the scheduler. This happens for like:

```
let obs = new Observable(observer => {
  throw "Crud"
})

obs.subscribe(next => {}) // Raises an exception with the scheduler
obs.subscribe(next => {}, error => {}) // Doesn't

Comment (zporter) [reply packages/microcosm/src/observable.js]: worth adding this as a comment in the code?

Comment (nhunzaker) [reply packages/microcosm/src/observable.js]: Certainly.

File: packages/microcosm/test/unit/domain/register.test.js [hunk 2]
```diff
@@ -63,4 +63,38 @@ describe('Domain::register', function() {
 
     expect(repo.state.test).toBe('test')
   })
+
+  it('should have state before an action resolves', async () => {
+    expect.assertions(2)
```
zporter: what does this line do?

nhunzaker: Fails the test if 2 assertions aren't given. This is useful for making sure code in `try/catch` executes, or that callbacks don't duplicatively fire (particularly nice for catching cleanup issues)

zporter: neat! thanks for the explanation!

efatsi:
👍 thanks for all the inline comments on the more obscure things

nhunzaker:
Thanks, guys! @zporter I've added a note about that case, as you requested.

PR #519, opened 2018-06-18
Pass next when completing with a payload. Improve efficiency

nhunzaker:
**What**

As a part of the Microcosm 13 update, actions are moving away from a custom data type and embrace the [Observable](https://github.com/tc39/proposal-observable) standard interface. Unfortunately, that causes a few API incompatibilities that I'm trying to smooth over before a full release of 13.x.

In Microcosm 12.x, you can complete an action with a payload like:

```javascript
action.resolve('stuff')
```

As far as I understand, with [Observables](https://github.com/tc39/proposal-observable), the correct way to accomplish this is something like:

```javascript
observer.next('stuff')
observer.complete()
```

While I've respected that for Observables themselves, this PR updates the Subject API to allow passing a payload to `complete`. This improves ergonomics and compatibility with usage in Microcosm 12.x.

**Other related work**

The drawback to this approach is that an extra action status is dispatched. I have alleviated that by adding a `filter` method to the Observable prototype that Domains use to avoid dispatch if an action update has no impact on their state. Essentially:

```javascript
// In a Domain
// History is a tree of actions that reprocess in a consistent order as they change
history 
  .filter(action => {
    // Can I process this action at all?
    return this.shouldListenTo(action)
  })
  .subscribe(action => {
    // Update the domain's state value
    this.dispatch(action)
  })
```

This actually ended up being a greater efficiency update than I anticipated, allowing me to unskip a few tests. Pretty neat. I've been wanting to add `filter` to the core API, but was wary of doing so until I could dogfood it internally. Now I can!


packages/microcosm/test/unit/observable.test.js
```diff
@@ -309,4 +309,62 @@ describe('Observable', function() {
...
+
+  describe('::filter', () => {
+    it('operates like array.filter over a stream of values', () => {
+      let isOdd = n => n % 2
+      let double = Observable.of(0, 1, 2, 3, 4).filter(isOdd)
+      let answers = []
+
+      double.subscribe(value => answers.push(value))
+
+      scheduler().flush()
```
zporter: is `double` `[1, 3]` before the subscribe line? is it worth asserting that `answers` is empty until after the scheduler has been flushed?

nhunzaker: It does not because it is lazy. The sequence doesn't process until you call `subscribe`. 

Seems like a good clarification anyway though, and a nice confirmation should I refactor this later. I'll add it.

nhunzaker: Did this in a5ee6303

nhunzaker:
@zporter Thanks! Honestly, as stop and go as this has been, they've been super helpful for me too.

nhunzaker:
Thanks again, @zporter!

PR #521, opened 2018-06-26
Add confirm prop to ActionButton and ActionForm

nhunzaker:
**What**

This PR adds a `confirm` prop to ActionButton and ActionForm that allows users to bail out of action dispatch if the result is false. Like:

```javascript
function askUser(value, event) {
  return confirm("Are you sure you want to do this?")
}

function ConfirmButton(props) {
  return (
    <ActionButton action={deleteItem} confirm={askUser}>
      Delete Item
    </ActionButton>
  )
}
```

**Why**

ActionButton doesn't have a great way to confirm actions like deleting a record. For symmetry, I also added this functionality to ActionForm.

---

**Note:** This is against 12.x, our stable release. I'll eventually move this over to 13.x too.


tommymarshall:
Very cool!

PR #523, opened 2018-06-27
[microcosm] Add flow types for scheduler

nhunzaker:
**What**

Adds [flow type](https://flow.org/en/) definitions for Microcosm's scheduler.

**Why**

The scheduler is a critical piece of Microcosm 13, and it didn't have any flow type coverage.

Adding flow types uncovered a bug where Observable was accessing the scheduler's error handler from the wrong namespace. This PR additionally adds test coverage for that case.

### DIFF
--- a/packages/microcosm/src/observable.js
+++ b/packages/microcosm/src/observable.js
@@ -102,7 +102,7 @@ export class Subscription implements Unsubscribable {
       // obs.subscribe(next => {}) // Raises an exception with the scheduler
       // obs.subscribe(next => {}, error => {}) // Doesn't raise
       if (observer.error === noop) {
-        scheduler().raise(error)
+        scheduler()._raise(error)
       }
     }
 
--- a/packages/microcosm/test/unit/observable.test.js
+++ b/packages/microcosm/test/unit/observable.test.js
@@ -371,7 +371,7 @@ describe('Observable', function() {
     })
   })
 
-  describe('scheduler error handling', () => {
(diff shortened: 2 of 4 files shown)


packages/microcosm/src/observable.js
```diff
@@ -102,7 +102,7 @@ export class Subscription implements Unsubscribable {
       // obs.subscribe(next => {}) // Raises an exception with the scheduler
       // obs.subscribe(next => {}, error => {}) // Doesn't raise
       if (observer.error === noop) {
-        scheduler().raise(error)
+        scheduler()._raise(error)
```
nhunzaker: 🤦‍♂️

packages/microcosm/test/unit/observable.test.js
```diff
@@ -370,4 +370,27 @@ describe('Observable', function() {
       )
     })
   })
+
+  describe('scheduler error handling', () => {
+    let raise = null
+
+    beforeEach(() => {
+      raise = scheduler()._raise
+      scheduler()._raise = jest.fn()
```
nhunzaker: I don't like patching the scheduler as a global singleton (each window gets a unique instance). I wonder how I could write this such that I could control what scheduler the Observable uses.

nhunzaker: Maybe something like `new Observable(callback, scheduler)`

nhunzaker:
@leobauza I think your eyes are great :)

PR #524, opened 2018-06-29
Handle promise errors using second argument of then()

nhunzaker:
**What**

This commit changes the way Subjects and Microcosm action resolution handle promises such that the error state is passed into the second argument of `promise.then()`.

This prevents a case where callbacks following the success case could raise an exception, but it would be trapped by the `.catch()` API call. So errors would "disappear" in the following case:

```javascript
const repo = new Microcosm()

repo.addDomain('test', {
  register() {
    return {
      action: () => {
        throw 'Oops'
      }
    }
  }
})

await repo.push('action') // => where did the error go?
```


nhunzaker:
Thanks, @dce !

